PR #53, opened 2020-01-31
[UPDATE] add imageUrl to notifiers

Bart274:
Equivalent of https://github.com/evilhero/mylar/pull/2409


evilhero:
Merged into python3-dev, but used the updated calls so that it no longer requests the image from CV, but extracts the downloaded cover into memory and sends the image (also added config options for both pushover & telegram).

@Bart274 you did the ground work on this, I just got the image loading into memory working (which is something myself and @barbequesauce have been tinkering with for awhile to no avail up until now). Cheers for this!

PR #61, opened 2020-02-02
[FIX] platform independent devnull location + bug in sending covers to Telegram

Bart274:
Without this fix, the git version check fails on Windows

### DIFF
--- a/mylar/notifiers.py
+++ b/mylar/notifiers.py
@@ -350,15 +350,23 @@ class TELEGRAM:
     def notify(self, message, imageFile=None):
         payload = {'chat_id': self.userid, 'text': message}
         sendMethod = "sendMessage"
+        files = None
 
         if imageFile:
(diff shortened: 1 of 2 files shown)


barbequesauce:
TIL os.devnull was a thing.

Bart274:
@barbequesauce I didn’t knew it either until I started looking into the issue I had to run the py3 version and saw it was already used in this project here: https://github.com/mylar3/mylar3/blob/05652beef1172b58f98624c8ea34d52d4f881c12/Mylar.py#L34

Bart274:
@evilhero this also fixes sending covers to Telegram which was broken since you changed the code to the extracted covers. I'll make branches in the future to prevent PR's for multiple fixes

evilhero:
All good in development (haven't tested telegram obviously)! Cheers!

PR #79, opened 2020-02-11
Add max version of natsort

Bart274:


barbequesauce:
Concerns with newer versions of natsort as CT pinned the version.

Did you test this change in your fork before submitting?

barbequesauce:
@Bart274 we set a lower max requirement in a previous commit... thanks for the research.

PR #104, opened 2020-02-24
fix turnitoff error

ajslater:
Fix calling turnitoff() before its declared.
Same as the patch from mylar2


barbequesauce:
Thanks!

PR #105, opened 2020-02-24
implement torznab tester

ajslater:
torznab test implementation for mylar3

torznab test was only partially implemented on the front end and never implemented on the back end, which of course made the test button fail as noticed in #2035

The partial front end torznab test implementation did not expose a Verify SSL button, but this PR does implement one on the front and back end.

### DIFF
--- a/data/interfaces/default/config.html
+++ b/data/interfaces/default/config.html
@@ -917,6 +917,10 @@
                                                     else:
                                                         torznab_enabled = ""
 
+                                                    if torznab[2] == '1' or torznab[2] == 1:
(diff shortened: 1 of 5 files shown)


ajslater:
mylar3 now claims its querying a torznab host when it searches, but idk if it's speaking torznab correctly or what. I haven't had any successful search returns yet.

barbequesauce:
OK - let's make sure it works before merging :)

ajslater:
This has been tested with Jackett & deluge and works reliably.
It works much better when #116 is applied because then it can handle magnet links as returned by the torznab host.

PR #106, opened 2020-02-24
fix import to use metadata key

ajslater:
Import dies on a key error trying to access issueinfo[0].
accessing issueinfo['metadata'] fixes it.

### DIFF
--- a/mylar/librarysync.py
+++ b/mylar/librarysync.py
@@ -21,6 +21,7 @@ import glob
 import re
 import shutil
 import random
+import traceback
 
 import mylar
 from mylar import db, logger, helpers, importer, updater, filechecker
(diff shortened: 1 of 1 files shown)


evilhero:
This looks good - I had changed the return dict to separate meta-data so that it could be used in various other modules. Completely forgot the import part since I have that already done on my local repo, but haven't merged anything as it's still a WIP. 

When I merge my changes, the only update will be that I've set ```meta_data = issueinfo['metadata']``` immediately after the else statement on  line 290, which then just changes the subsequent calls to just ```meta_data['series']``` instead of ```issueinfo['meta_data']['series']```. It doesn't have any impact other than just that, so it shouldn't affect anything else going forward. 

I'm ok to merge this ;)

barbequesauce:
thanks!

PR #107, opened 2020-02-24
cast maxyear as an int for range()

ajslater:
Occasionally maxyear ends up as a float, probably a result of an automagic string conversion. Casting it to int makes range() not throw an exception.
I ran into it with a value of 2004.0 which casts nicely


barbequesauce:
thanks!

PR #115, opened 2020-02-25
Fix crash with null issue number.

ajslater:
Sometimes issue number is None, which makes the regex crash.


barbequesauce:
Thanks!

PR #116, opened 2020-02-25
fix for magnet links, especially from torznab

ajslater:
Fix handling of magnet links, especially from torzab search.
If the url is a magnet link or if the url redirects to a magnet link, don't write a file with the torrent contents, since there aren't any. Just pass the magnet string along to the downloader.

This PR is a sibling to #105


barbequesauce:
Thanks!

PR #119, opened 2020-02-27
Auto proxy support

ajslater:
The way many ajax buttons are implemented in mylar is through server side redirects. This is nice and easy but if the server doesn't peep at the headers to see that SSL is terminated somewhere else, it redirects to http instead of https  and causes a UI error.
The button "worked" it just doesn't redirect to the proper success notification and appears to have failed until you do a manual client side refresh.

This fixes about a hundred apparent bugs for people using SSL termination.

### DIFF
--- a/mylar/webstart.py
+++ b/mylar/webstart.py
@@ -74,8 +74,8 @@ def initialize(options):
 
     conf = {
         '/': {
-            'tools.staticdir.root': os.path.join(mylar.PROG_DIR, 'data')
-            #'tools.proxy.on': True  # pay attention to X-Forwarded-Proto header
+            'tools.staticdir.root': os.path.join(mylar.PROG_DIR, 'data'),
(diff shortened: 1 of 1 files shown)


barbequesauce:
@evilhero this must have been commented out for a reason...?

ajslater:
Git blame says it entered the code as a part of a big change already commented out.
https://github.com/mylar3/mylar3/commit/55eedb186917018c4022d9be0624f28395ccd6d9

evilhero:
Was probably due to the inclusion of proxy chains, but I can't honestly remember. Not sure if this retains chain usage or not as I don't run any chains, and I can't recall who was testing it for us.

barbequesauce:
Works as advertised. Thanks!

ajslater:
Woo! thank you!

PR #120, opened 2020-02-28
Update requirements

ajslater:
The major changes are:

* Add html5lib, which is explicitly referenced by mylar
https://github.com/mylar3/mylar3/blob/05652beef1172b58f98624c8ea34d52d4f881c12/mylar/wwt.py#L56
https://github.com/mylar3/mylar3/blob/05652beef1172b58f98624c8ea34d52d4f881c12/mylar/wwt.py#L80
* Separate the comictagger dependancies and keep them as absolute copies of those specified by comictagger.
* Change the mylar specific dependancies to use =~ operator instead of >= just so a major incompatibly version change doesn't break us.
* Sort alphabetically.

Wanted to do this before making a Dockerfile

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 # `pip install -r requirements.txt` before you can run this.
 # this might not be exhaustive, as this is still a WIP.
 #### ESSENTIAL LIBRARIES FOR MAIN FUNCTIONALITY ####
-## for mylar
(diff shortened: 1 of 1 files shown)


ajslater:
shelve this for now. was mostly a prereq for a Dockerfile because of html5lib which is actually not required and so its reason for being has evaporated.

PR #125, opened 2020-02-29
Fix regex parsing crash parsing search results

ajslater:
the argument 'i' is never a regex and yet sometimes it contains regex special characters like parentheses. In this case it usually crashes because the parens are unmatched, but there's a chance it could behave strangely by interpreting matched parens or other characters as a regex instruction.

Change this to a string replace operation to fix the crash or undefined behavior

Currently testing.


barbequesauce:
Easier to read too. Thanks!

PR #129, opened 2020-02-29
pidfile tweaks

btxgit:
1. Add traditional PID-based pidfile: I saw the pidfile was being created with the word "pid" in it, rather than the parent PID of the daemonized process, which is the tradition. I think maybe pidfile was simply used as a lockfile, but there's no reason to limit ourselves to that since the daemonize code is limited to non-win32.

2. Add a check to determine if an existing pidfile is stale (meaning it represents a process that no longer is running).  This function does nothing unless you're on linux with a /proc fs - in which case it will check  if the given pid exists currently.  If not, we know it's a stale PID file and we return True.  If it DOES exist, it checks the cmdline to make sure the commandline is a python process.  If python isn't in the cmdline, we know it's a stale pidfile.  If it does, we assume it's not stale.

There's room for improvement in the cmdline checking aspect.

### DIFF
--- a/Mylar.py
+++ b/Mylar.py
@@ -37,6 +37,33 @@ if ( sys.platform == 'win32' and sys.executable.split( '\\' )[-1] == 'pythonw.ex
 def handler_sigterm(signum, frame):
     mylar.SIGNAL = 'shutdown'
 
+def check_stale_pidfile(pidfile):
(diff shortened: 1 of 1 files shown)


barbequesauce:
Makes sense to me - but I’d still ask @evilhero to take a gander...

barbequesauce:
Thank you sir!

PR #130, opened 2020-02-29
Fix restart queue empty

ajslater:
Simple fix to prevent that.


barbequesauce:
Nice catch, thanks!

PR #131, opened 2020-02-29
Fix empty issname crash

ajslater:
Occasionally len(issname) can be 0 which causes a crash.

This PR indents a large block of code to prevent all the operations the extract data from issname. I *think* that's the right thing to do, but someone with more context could propose a better solution.

### DIFF
--- a/mylar/torrent/clients/deluge.py
+++ b/mylar/torrent/clients/deluge.py
@@ -126,21 +126,22 @@ class TorrentClient(object):
 
 
                 #Check if torrent already added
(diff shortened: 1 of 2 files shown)


barbequesauce:
How has this one been tested? That’s a lot of change...

ajslater:
Just by me running the searcher and not running into it again. Indenting that much code is admittedly awkward.

barbequesauce:
Can you close this PR and try again without re-merging the branch back in please?

PR #132, opened 2020-02-29
Revert "Fix restart queue empty"

barbequesauce:
Reverts mylar3/mylar3#130

PR #136, opened 2020-02-29
Fix restart queue empty

ajslater:
Simple fix to prevent that.

Replaces #130 (reverted by #132) as this one is against the correct dev branch.


barbequesauce:
Can you close this PR and try again without re-merging the branch back in please?

PR #137, opened 2020-02-29
pidfile normalcy

btxgit:
1. Make pidfile actually use the PPID rather than the word "pid"
2. Add stale pid file detection which works like:
- Is paltform linux w/proc?  No?  It's not stale
- Read the pid from pidfile.  Does it exist on this system?  No?  It's stale
- Read if /proc/{pid}/cmdline, converting '\0' to spaces.  If python in the cmdline?  If no, then it's stale.  Otherwise we assume it's not.


barbequesauce:
thanks again!

PR #140, opened 2020-03-01
fix restart

ajslater:
Simple fix to prevent that.

Replaces #130, #136, #139

Manually rebased


barbequesauce:
Thanks!

PR #232, opened 2020-03-28
Fix crash if Mylar sees book that's a symlink to a nonexistant file

btxgit:
I had a few broken links in my comic share that were causing Mylar to stop processing my comics.

To reproduce:

cd /path/to/comics
ln -s /path/to/nothing

[python]
for fn in os.listdir('/path/to/comics'):
     if fn == 'nothing':
         exist = "Yes" if os.path.exists(fn) else "No"
         print(f'''Filename {fn} exists: {exist}''')

Without my patch, it chokes on the getsize call.


barbequesauce:
Thanks!

PR #235, opened 2020-03-29
include_comicrack_style_issue_id setting to add CVDBXX Issue ID to Notes

SenorSmartyPants:
Add an option to CT settings [comicvine] section include_comicrack_style_issue_id. Default value False. mylar3 behavior does not change when value is False.

When set to true, Notes field will include [CVDBXXXXX]. XXXX is issue ID. This is the format that the ComicVine Scraper for ComicRack reads. It will allow CR to automatically pick up the issue ID and not prompt user on first scrape in CR.

example output when set to true.

Tagged with the ninjas.walk.alone fork of ComicTagger 1.25.2 using info from Comic Vine on 2020-03-28 13:00:47. [Issue ID 721150] [CVDB721150]

### DIFF
--- a/lib/comictaggerlib/comicvinetalker.py
+++ b/lib/comictaggerlib/comicvinetalker.py
@@ -513,7 +513,11 @@ class ComicVineTalker(QObject):
         if settings.use_series_start_as_volume:
             metadata.volume = volume_results['start_year']
 
-        metadata.notes = "Tagged with the {0} fork of ComicTagger {1} using info from Comic Vine on {2}.  [Issue ID {3}]".format(
+        notes_format = "Tagged with the {0} fork of ComicTagger {1} using info from Comic Vine on {2}.  [Issue ID {3}]"
+        if settings.include_comicrack_style_issue_id:
+            notes_format += " [CVDB{3}]"
+
+        metadata.notes = notes_format.format(
             ctversion.fork,
             ctversion.version,
             datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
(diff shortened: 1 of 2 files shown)


evilhero:
This is really just adding the same exact information (the issueid) twice to the same exact field. I understand the reasoning behind it, however the implementation behind it makes no sense (```IssueID:``` vs ```CVDB```).

I think in a case like this it would be a simple matter for CVScraper to just look for ```Issue ID``` as it does for ```CVDB``` when the latter fails, and it would save users of either CVScraper or ComicTagger having to rescrape everything again. However I also know that the author of CVScraper has put it the "lights on cept for when the power goes out" kinda deal. Having any parser (regardless of which it is) be able to be read _both_ formats would be beneficial to everyone in acknowledging relevant sources other than just themselves. 

@cbanack, I know you're busy with life and everything (fellow Canuck, eh!) and from my quick code glance at yours it seems like you only check for the ```CVDB``` wording in one location (cvdb.py_parse.key_tag) - so adding something to allow for ```Issue ID``` should be quick and painless if I read the code correctly of course:
```
def _parse_key_tag(text_s):
   ''' ComicVine implementation of the identically named method in the db.py '''
   
   tag_found = re.search(r'(?i)CVDB(\d{1,})', text_s)
   if not tag_found:
      tag_found = re.search(r'(?i)Issue ID(\s\d{1,})', f)
      if not tag_found:
          tag_found = re.search(r'(?i)ComicVine.?\[(\d{1,})', text_s); # old format!
   return int(tag_found.group(1).lower()) if tag_found else None
```
Is this something that you could do easily from your side? I think it would be beneficial to everyone using any meta-tagging software to allow for different variations of critical information used for 
identification purposes.

SenorSmartyPants:
Closing this in favor of #236

PR #236, opened 2020-03-29
optional config.ini choice to use either [CVDBxxxx] or [Issue ID xxxxxx] when metatagging

evilhero:
This should supercede PR #235 (with thanks to @SenorSmartyPants)

Allows for optional configuration which will either use ```[CVDBxxxxxx]``` or ```[Issue ID xxxxxx]``` in the notes field when metatagging. If no option is selected, the default of ```[Issue ID xxxxxx]``` will be used when metatagging (as it is currently done).

**config.ini addtions**:
```
[metatagging]
ct_notes_format = Issue ID / CVDB (select only one)
``` 

From a CLI perspective this can also be achieved by using the ```--notes_format``` switch with either ```Issue ID``` or ```CVDB``` as switchable options.

Also bumped ct_version to ```1.3.3```


barbequesauce:
Nice compromise!

PR #246, opened 2020-04-10
Fix javascript errors on Import Results screen

eroth312:
This was due to some extraneous table rows that the datatables library was failing to process.  Without datatables loading properly, there was no sorting, paging, etc on the table of stuff to be imported.


evilhero:
First thanks for the PR. 

However, you did the PR against the Master branch - you need to do the PR against the python3-dev branch (which you committed to, but the PR is against the wrong branch). 

As well, your repo isn't up-to-date prior to doing the PR - thus all the extra commits that bog down the PR. Get your local repo up-to-date, and then do the PR against the right branch - basically the only item in the PR should be the changes that you are presenting, not everything else.

Do you also have a reference as to what the error was that was occurring? I don't see anything js related in the PR save for some <tr> removal outside the dom, which really shouldn't affect things from displaying.

eroth312:
Oh nuts - I had it locally on the right branch, but PR defaulted to master and I didn't catch that.   Let me close this and resubmit against the dev branch.

PR #249, opened 2020-04-11
Performance Tuning - speed up weekly pull check which was locking db

eroth312:
I was investigating a database that was consistently getting locked over time, and noticed the weekly pull query was taking an inordinate amount of execution time via a profiler.

I've replace the super slow INNER JOIN on ORs with a union against each table.   The performance difference against my database can be seen in the screen shots below.  This is a consistent change of about 30s down to less than 500ms.

A few notes, and I'm happy to dig into these more, if need be:
1.  I didn't find a use for the ComicPublisher field that's returned in the recordset.   I think it could be removed, but didn't want to overreach on my change.  I simply set it to NULL for the first set of data coming back from Annuals.
2.  Regardless which week I checked, I couldn't find any data coming back from my "annuals" table.   This might simply have to do with my setup, and I'd be happy to dig further if someone can put towards how to get proper data in here.   For what it's worth my table is not empty.

Screen Shot 2020-04-11 at 7 54 40 AM
Screen Shot 2020-04-11 at 7 55 19 AM


barbequesauce:
This looks good to me - and great work! - but there's places in Mylar i am hesitant to touch without @evilhero's agreement, and this is deep in the middle of one of them...

eroth312:
Thanks @barbequesauce.
No worries - I'm happy to bang away however to make sure it's solid.  I'm just trying to get my install running without issues. and this was a nasty one.  Next up is trying to figure out why CDH just stops working after awhile, but there's a ton of code to unravel there.

barbequesauce:
I think we will all be interested to see that answer!

evilhero:
More than happy to answer any questions as to the data collection methods - especially since you paved the way with enhancing the sql query ;)

I'll try and answer these in the order you asked, but pay no attention if I wander from one to the other.
1. The ComicPublisher field being used in the recordset returned at that point in the code isn't really being used from what I can remember. Initially I had put it in to be returned so that it could help verify the issue on the pullist by comparing the publisher on the pull to the publisher on the watchlist. But since we moved away from the old method of the pull having to compare based on series names alone (now we match based on the actual ID's and fallback to name matching if all else fails), the ComicPublisher could probably be removed safely from the query.

2. The annuals table is the table that's used when annual-integration is enabled. So the Issues table is used for normal issues, regardless of if annual-integration is enabled or not. If annual-integration is not enabled, then annuals are considered as separate volumes and are dropped into the Issue table as per normal. Now the annuals table is only used to store annuals when the annual-integration is enabled. This was done for a bunch of reasons, but the main reason being that it was a quick way to distinguish non-annuals from annuals regardless of if the flag was enabled or not in the configuration. So if no annuals are present on the weekly pull-list, and you do not have annual-integration enabled, you wouldn't see any data coming back from the annuals table, as both items have to present in order for data to get pulled.

As far as the SQL change you made, I do have two questions just for clarification: 
- how are you testing the speed differences between the two (ie. what are doing in the GUI and then testing the speed against - like a ```Recreate Pull-list``` option ?)
- even as per the old SQL method, my time was never 30s to retrieve the listing - however I also probably have a fairly small dataset in relation to others, so I was wondering how many items do you have on your watchlist?

I'm just testing this against my local repo now, but barring anything blowing up I don't see any reason why this wouldn't be able to be merged as is.

eroth312:
@evilhero I'm probably just being difficult on the app at the moment, since i was migrating from one organization method to mylar3, so importing 100s of things at a time.   Since this query runs in the background, my assumption is that it was colliding with the repeated writes to the comics table and totally exaggerating the problem, which was causing me to get into a database locked state.  Only recourse was restarting the whole app.    At that point, i started digging in via py-spy to figure out what is going on, which pointed to this particular query with a massive percentage of the runtime.

You're right - I'm set up to keep annuals as their own volumes.  Right or wrong, i guess it's just how it got pulled in and sort of makes more sense to me since it matches comicvine.  I was testing the query with 2020 week 4, since it was the most recent annual i could find.   It worked properly in my setup.

That said - i pulled the db out and to my laptop, where you see the raw query times against the db in my screenshots above.  That is only those queries running, with nothing else happening, so it's clearly not amazing.

To answer your question, I've got 902 wanted and 141 snatched at the moment.     My next adventure is figuring out why those are stuck in snatched.   I've also seen the "A search is currently in progress....queueing this up again to try in a bit." message throughout my log, which is strange, since it hasn't found a result in over 12 hrs.   Related or not, it seems wacky, and I might try to unravel that too, but i need to learn how search queuing and cdh work first.

evilhero:
Yeah the importer is currently going thru a rebase I guess you'd call it - I've been working on it for the past week or so (obviously more difficult to spend hrs coding during this time, for those of us with young children). I have it redone currently at the moment to segregate by series folders or by individual parsing so that the grouping is correct. Also addressed story-arc naming being imported (where the reading sequence comes first in a filename, the previous importer failed to read them correctly), and even added some pretty images and made it dynamically paginate. Of course in all of this, most of the code has to be rewritten to accommodate the new structures, and in doing so I'm finding little glitches that I'm trying to fix at the same time.

But yeah, doing an import would collide with anything else that might be going on in the background (or attempting to) - like the pull, the auto/manual searches, even post-processing I would guess. We've tried to try to gracefully pause/wait in some cases, but obviously we aren't catching them all.

There also is no right or wrong way to collect annuals - some users want to see all the annual volumes listed as part of the given series volume, others want it as individual volumes (like CV). It's more personal preference, but the option has always been there as it's something unique that only Mylar does atm.

I was inquiring more about the number of series you're watching (on the main/home page) - as that's really what the SQL enhancement is querying against.

The stuck in snatched thing has been an ongoing thing lately with CDH - I was able to track a portion of it down to the history reference that is being done by CDH (it queries the history of the download client in order to see if there's a unique match), but haven't been able to revisit it cause there's just not enough time in the day what with the other stuff that I'm currently working on. The ```search is currently in progress``` notif is probably just an errant logger line in the wrong place due to how the queues are - but it could also be related to a stalled search queue that errored out and is stuck in limbo. Hard to tell as every situation seems to be more unique than the last...

If you'd like to talk more about this kinda stuff - or hammer some things out - you should head onto irc where there are those of us that work on the dev stuff hang out. Might make things easier to communicate/ask questions...

PR #251, opened 2020-04-12
Eliminate endless loop in sabnzb queue

eroth312:
clean up some issues in the nzb queue monitoring which:
1.  Fix the broken reference in the log line within helpers.py which was breaking things, and forcing the re-queueing to fail
2.  Fix the endless loop:
In this case, if an nzb was no longer found in the queue within sab, it'd be loop endlessly looking for it.   Solved by adding a quick check to see if the nzb we're looking for exists in the result set at all, and if not, return with a new status and log it.  Eliminating the while loop altogether seems fine, as it was only ever looping through the same result from sabnzbd history.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -3160,8 +3160,10 @@ def nzb_monitor(queue):
 
             if any([nzstat['status'] == 'file not found', nzstat['status'] == 'double-pp']):
                 logger.warn('Unable to complete post-processing call due to not finding file in the location provided. [%s]' % item)
+            elif nzstat['status'] == 'nzb removed':
+                logger.warn('NZB seems to have been removed from queue: %s' % item['nzo_id'])
             elif nzstat['status'] is False:
-                logger.info('Could not find NZBID %s in the downloader\'s queue. I will requeue this item for post-processing...' % item['NZBID'])
+                logger.info('Download %s failed. Requeue NZB to check later...' % item['nzo_id'])
                 time.sleep(5)
                 mylar.NZB_QUEUE.put(item)
             elif nzstat['status'] is True:
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -3158,12 +3158,13 @@ def nzb_monitor(queue):
                 logger.warn('There are no NZB Completed Download handlers enabled. Not sending item to completed download handling...')
                 break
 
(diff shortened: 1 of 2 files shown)


mylar/helpers.py
```diff
@@ -3160,8 +3160,10 @@ def nzb_monitor(queue):
 
             if any([nzstat['status'] == 'file not found', nzstat['status'] == 'double-pp']):
                 logger.warn('Unable to complete post-processing call due to not finding file in the location provided. [%s]' % item)
+            elif nzstat['status'] == 'nzb removed':
+                logger.warn('NZB seems to have been removed from queue: %s' % item['nzo_id'])
             elif nzstat['status'] is False:
-                logger.info('Could not find NZBID %s in the downloader\'s queue. I will requeue this item for post-processing...' % item['NZBID'])
```
evilhero, line 3164: We have to keep the reference to NZBID, as that's the return from nzbget.py (the nzbget monitor) - it uses NZBID, whereas SAB uses the nzo_id. So unfortunately leaving it as in the PR will break nzbget usage.

mylar/sabnzbd.py
```diff
@@ -112,55 +112,64 @@ def historycheck(self, nzbinfo):
...
-                                                 'location': os.path.abspath(os.path.join(hq['storage'], os.pardir)),
-                                                 'failed':   True,
-                                                 'issueid':  sendresponse['issueid'],
-                                                 'comicid':  sendresponse['comicid'],
-                                                 'apicall':  True,
-                                                 'ddl':      False}
-                                break
+
+        nzo_exists = False
+        for hq in histqueue['slots']:
```
evilhero, line 117: Not sure about the extra for/loop here as it seems kinda redundant, but maybe if we remove the 2nd for/loop since we already have the matching hq value, store that the hq value here (~line 119.. ```hqdata=hq```) and just run thru the subsequent if/elseif's referencing hqdata instead of hq, as there would be no for/loop.

eroth312: Yeah - good point.  let me refactor that a bit to remove the double loop.   That was just leftover from removing the while, and i hadn't paid too much attention to what was inside.

eroth312:
@evilhero Both of your comments should be addressed in that last commit.    Let me know if we want to adjust anything else.

PR #258, opened 2020-04-22
Update search.py

samsonious:
fix typo febug -> fdebug


mylar3:
Thanks for the PR - however, there are 2 things to keep in mind:
- this has already been fixed in python3-dev, but has yet to merge with master.
- we never push any PR's to the master branch, always send any PR's to the python3-dev.

So we won't merge this, and I'll close it. But thanks regardless for the PR - cheers!

PR #268, opened 2020-04-24
Adding Discord notifier

hubbcaps:
Adding Discord notifier


mylar/PostProcessor.py
```diff
@@ -2844,6 +2844,10 @@ def sendnotify(self, series, issueyear, issuenumOG, annchk, module, imageFile):
             slack = notifiers.SLACK()
             slack.notify("Download and Postprocessing completed", prline2, module=module)
 
+        if mylar.CONFIG.DISCORD_ENABLED:
+            discord = notifiers.SLACK()
```
evilhero, line 2848: Think this is supposed to be ```notifiers.DISCORD()``` and not ```notifies.SLACK()```

hubbcaps: Whoops, missed that one. Should be good now.

PR #288, opened 2020-05-17
fix a version checking crash

ajslater:
Running mylar3 on python3-dev branch it looks like sometimes this `git` json array is empty when it returns from github.
```
  File "/app/mylar/Mylar.py", line 386, in <module>
    main()
  File "/app/mylar/Mylar.py", line 356, in main
    versioncheck.versionload()
  File "/app/mylar/mylar/versioncheck.py", line 364, in versionload
    version_info = getVersion()
  File "/app/mylar/mylar/versioncheck.py", line 114, in getVersion
    if git[0]['name'] is not None:
```

### DIFF
--- a/mylar/versioncheck.py
+++ b/mylar/versioncheck.py
@@ -110,23 +110,22 @@ def getVersion():
             except Exception as e:
                 logger.warn('[ERROR] %s' % e)
                 pass
-            else:
-                if git[0]['name'] is not None:
-                    for x in git:
-                        if x['name'] == output[:opp]:
-                            current_version_name = x['name']
-                            cur_commit_hash = x['commit']['sha']
-                            break
-                    logger.info('version_name: %s' % current_version_name)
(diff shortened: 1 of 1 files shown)


evilhero:
Is this something that can be reproduced easily? While the patch will fix the None / crash, what does that do for the rest of the flow in regards to the git value it's trying to determine ?

Is there a way you can throw a logger line around line 114 of versioncheck.py and check to see what the value of the ```git``` variable is as well as the value of ```url2``` is? It shouldn't be returning an empty value for name since it's pulling it down via the tags api of github. Maybe if you're using an outdated fork, I can see it returning nothing in some cases (you'd have to set your mylar.CONFIG.GIT_USER to your personal fork though for that to happen), but I'm struggling to see how it would occur otherwise.

ajslater:
Thanks for prompting me to look further into this. This bug was caused by an outdated config file:
```
[Git]                                                                       
git_user = evilhero
```
When the new git_user for the new repo should be `mylar3`

I don't think my patch is needed.

Issue #329, opened 2020-06-01
Mako Runtime Error


burnshroom:
When selecting The Flash (2011) (comicvine ID: 4050-43018, an error occurs with a tab titled "Mako Runtime Error"

The page displays this:

Error !
KeyError: 'Issuename'
<td id="aint_issuenumber">${annual_Number}</td> <td id="aissuenumber">${annual['Issue_Number']}</td> <% if annual['Issuename'] is not None: if len(annual['IssueName']) > 70: aissuename = "%s..." % annual['IssueName'][:70] else: aissuename = annual['IssueName'] try:
/mnt/mylar3/data/interfaces/carbon/comicdetails.html, line 587:
if annual['Issuename'] is not None:
/mnt/mylar3/data/interfaces/carbon/base.html, line 87:
${next.body()}
/home/shroom/.local/lib/python3.8/site-packages/mako/runtime.py, line 947:
callable_(context, *args, **kwargs)
/home/shroom/.local/lib/python3.8/site-packages/mako/runtime.py, line 920:
_exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
/home/shroom/.local/lib/python3.8/site-packages/mako/runtime.py, line 878:
_render_context(
/home/shroom/.local/lib/python3.8/site-packages/mako/template.py, line 476:
return runtime._render(self, self.callable_, args, data)
/mnt/mylar3/mylar/webserve.py, line 56:
return template.render(http_root=mylar.CONFIG.HTTP_ROOT, **kwargs)

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Branch / commit: python3-dev / f1cc119

barbequesauce:
@burnshroom give that a shot please

burnshroom:
The issue remains.

Also affecting Fantastic Four volume 1 (comicvine ID: 4050-2045)

barbequesauce:
Which interface?

burnshroom:
Carbon.  Both open fine in default interface.

barbequesauce:
make sure you're current, then give it one more shot please

burnshroom:
Current and tested... will close as it seems to be fixed. 

Thanks.

PR #332, opened 2020-06-03
FIX: #329 (carbon-flavored)

barbequesauce:


evilhero:
This should go to python3-dev first else we'll be outta whack on our favourite git system again - my fault we missed it.

Issue #337, opened 2020-06-07
Traceback on many issues when doing Import


bguidotti:
Hi. I'm on the development branch of Mylar3, running on Win10 with Python 3.7.7.
I'm trying to do a mass import of a bunch of files that a duplicate removal program left hanging out in a miscellaneous directory. In other words, they used to be in the correct series folders, but now are in a single folder.
The importer works on some files, but pukes out pretty quickly for me if I try to set it and walk away.
I'm hoping some of my filenames will help improve it.

I turned on debug logging and did a carepackage after I tried an import again.
I can get that file to you through Google Drive.

**Here is a sample:**
06-Jun-2020 22:29:09 - DEBUG :: mylar.preSearchit.4806 : Thread-24 : search_values: [{'SRID': '347271', 'comicid': '94624'}]/{'Series': 'Enchanted Tiki Room', 'results': 2, 'publisher': 'Marvel', 'haveit': {'comicid': '94624', 'status': 'Active'}, 'name': 'Enchanted Tiki Room', 'deck': 'None', 'url': 'https://comicvine.gamespot.com/enchanted-tiki-room/4050-94624/', 'description': '<p>Five issue mini-series.</p><p>Collected in <a href="https://comicvine.gamespot.com/enchanted-tiki-room/4050-100693/" data-ref-id="4050-100693">Enchanted Tiki Room</a>.</p>', 'comicimage': 'https://comicvine1.cbsistatic.com/uploads/scale_large/6/67663/5467806-01.jpg', 'issues': '5', 'ogcname': 'Enchanted Tiki Room', 'comicyear': '2016'}
06-Jun-2020 22:29:09 - ERROR :: mylar.excepthook.315 : Thread-24 : Uncaught exception: Traceback (most recent call last):
  File "C:\mylar3\mylar\logger.py", line 337, in new_run
    old_run(*args, **kwargs)
  File "C:\Users\Bruce\AppData\Local\Programs\Python\Python37\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\mylar3\mylar\webserve.py", line 4807, in preSearchit
    myDB.upsert("searchresults", nVal, cVal)
  File "C:\mylar3\mylar\db.py", line 169, in upsert
    self.action(query, list(valueDict.values()) + list(keyDict.values()))
  File "C:\mylar3\mylar\db.py", line 128, in action
    sqlResult = self.connection.execute(query, args)
[...]

Issue #149, opened 2020-03-02
Better importer for large comic collection


ErenPhayte:
The current importer doesn't scale well, and if comicTagger fails, it fails and stops importing or the endpoint (markImports?action=massimport) timeout and the importing stops completely. I keep having to hit that endpoint every time so that it can mass import a few comics at a time which means this will take forever to get in.

Suggestion:

Move the importer to a Queue for processing. Allow for multiple jobs to run simultaneously (configuration) to speed up importing. Use the queue to recover from failed imports by pushing job back into the queue for later.

If there is a challenge with the table locking, then process all comics back into a queue after you have retrieved the metadata you needed for DB inserts. This will allow you to scale on the metadata reading part and preparing the SQL statements for import. Then have a queue process these in batch SQL inserts or 1 insert at a time but at least the heavy lifting is done prior to the inserts (unless you have dependencies or other complexities that might prevent this from scaling)

evilhero:
ComicTagger isn't used at all during importing. However they share some of the same modules which would be causing the lockup if it does happen. Mylar, as it currently stands at least, will not be able to import TPB, GN's due to not having an issue number which when the importer was built was a requirement. This will get updated, but not immediately.

Allowing multiple imports to run simultaneously will cause db locking and god knows what else - so know, unfortunately that's not an option. The Queue thing is a possibility, as we use that for various other functions (ddl downloads, post-processing, searching). However the importer would have to be rewritten to accommodate the queue usage and those of us on the dev team are busy patching due to the merge, not enhancing atm. Will mark this as an enhancement request until such time as we have the time to look more into this

Also, speed of importing will be limited by your collection. If you enable parsing metadata, it has to load in each file and parse the metadata, store it, and then move on to the next one, etc. If you have 120,000 comics, you're gonna be awhile regardless. If speed is crucial, and your files are named sanely - you could possibly get away with turning off the parse metadata option, however then it has to parse each and every filename regardless and there's your time buffer shot again.

ErenPhayte:
Just want to make sure we talking about the same thing. 
Currently its this link /markImports?action=massimport that times out.

evilhero:
Any import action does not involve ComicTagger usage currently. The mass import queues up all the queued up files and attempts to auto-add them to your watchlist.

If you have a debug log to show what's going on if you have questions, it's best to provide that instead of just me trying to be hypothetical based on what you're asking.

ErenPhayte:
[...]

PR #338, opened 2020-06-07
FIX:(#337)(#149) Partial fix for stalled importer throwing a DB exception

evilhero:
This is a partial fix for the importer stalling and throwing a DB exception due to datatype. There are upcoming updates that will address #337 and #149 fully, but this is the simplest of ones to patch to get the importer working.

This will allow it to continue in most cases, but subsequent errors are a possibility as a result. Note that this *might* also fix the problem of manual intervention returning a blank page in some cases.


bguidotti:
@evilhero with the quickness! Thanks, guys!

PR #342, opened 2020-06-07
Icon fix and minor tweaks

sukhysall:
1. Fixed the arrow icons in the Weekly pulllist to an all white colour.
2. Fixed the History icon. Inverted now to match the other icons.
3. Decreased the font size in the Options screen.
4. Centered the History tabular results.


sukhysall:
Ahhh!!!

I thought I had uploaded to data/interfaces/carbon/images. Doh!

Okie Dokie. I'm gonna close this off then re-do tomorrow PM. Didn't realise the time.

sukhysall:
@barbequesauce thanks for the instructions. 

Just familiarising myself with what I need to change and where. Looks slightly daunting.....and then I have another doh moment when I realise that any cocksup will not affect the main fork!!

This is embarrasing. Thanks again.

barbequesauce:
All good! We created this dev branch because we’re messing with some core functionality- we can break whatever we want and only impact you, me, and evilhero (and it’s easy enough to revert changes)

evilhero:
Yeah we did it to minimise the amount of stuff being duplicated - that and we were always forgetting one of the interfaces when we were updating, so hopefully this will fix some of those issues (to make it clear it was mainly our problem lol).

So if you don't touch any of the html files, then they live in the interfaces/default folder now. Everything in the default folder are considered our base files. Any file that you modify for your carbon template, would need to go into the respective js/css/images folder within the ```interfaces/carbon``` folder.  If you were to modify say comicdetails.html, then you would place the modified file in ```data/interfaces/carbon/comicdetails.html```.  

In doing this, Mylar will look for it in the given interface directory - if it can't find it, then it will just load up the base file(s) so that things aren't completely broken.

The only caveat in all of this, is that Mylar needs to know what we're changing around. So in the webserve.py module, and in the serve_template function you'll see the icons dict that we need to modify that looks like this (a portion of it): 
```'icon_gear': os.path.join(mylar.CONFIG.HTTP_ROOT, 'interfaces', 'carbon', 'images', 'icon_gear.png'),```

So adding say ```prev.png``` as an updated carbon theme image, but we're not touching the default interface one. So to the icon list we would need to add under the default icons (pay no attention to the icon_name (```prev.png```) - I'm just doing it for simplicity here:
``` 'prev.png': os.path.join(mylar.CONFIG.HTTP_ROOT, 'images', 'prev.gif')```

and then under the else statement (which is for the carbon interface):
``` 'prev.png': os.path.join(mylar.CONFIG.HTTP_ROOT, 'interfaces', 'carbon', 'images', 'prev.png')```

Then the final step (assuming you already put the new prev.png in the ```interfaces/carbon/images``` folder, is to modify the default template that uses it. As an example I'll use previous arrow in comicdetails.html:

```<a href="comicDetails?ComicID=${series['Previous']}"></a>```

and change it so that it can reference the icon dict we setup to serve out above:
```<a href="comicDetails?ComicID=${series['Previous']}"></a>```

Now it will reference ```prev.gif``` when using the default template, and use ```prev.png``` when using the carbon interface.

Hope that helps explain things abit - plus we're always around if you have any problems.

PR #355, opened 2020-06-21
Log the exception on search failure but allow other searches to run.

miracle152005:
searchforissue is now try catched and the searchlock is cleared in a finally


miracle152005:
@barbequesauce : wouldn't it then be better to not catch the exception, but just release the SEARCHLOCK in the finally block.
This would result in the expcetion bubbling up to the calling methods an being handled there but still release the searchlock for additional searches.
Or does this also raise concerns about the download or post-processing?

evilhero:
You have to have the exception in there. You don't have to log it, but it needs to be present. Releasing the searchlock does nothing if there's a traceback - the queue would still fail as it wouldn't get to that point without the try/except. 

The problem with throwing a try around it the way it's being done in the PR is that any exceptions that occur in a child process are not logged properly in the parent. So in your case, based on the previous issue you brought forth, the log would indicate the error, but nothing else. No traceback, no line references, nothing. So from a logging perspective, it's as much information as having mylar log in just INFO mode which is utterly useless when trying to fix anything as we wouldn't be able to reference the actual problem. 

I've cobbled something together that might fix some of these issues, along with the problems that we have with this PR.

miracle152005:
@evilhero: 
"The problem with throwing a try around it the way it's being done in the PR is that any exceptions that occur in a child process are not logged properly in the parent. So in your case, based on the previous issue you brought forth, the log would indicate the error, but nothing else. No traceback, no line references, nothing."

Currently I do not understand this part of your answer, most likely because my knowlege of python is limited (having done most of my development in java or c#), but isn't that the difference between logger.error and logger.exception? Just wanting to make sure I do not miss something about the logging of python.

At least my tests resulted in logger.error only displaying the error message itself but logger.exception showing the stacktrace.
While it is true, that it is differently formated, it should still be sufficient to analyze such issues.

This is a stack trace I am getting while catching the exception and writing it with logger.exception:
23-Jun-2020 16:08:57 - ERROR   :: mylar.searchforissue.1914 : SEARCH-QUEUE : int() argument must be a string, a bytes-like object or a number, not 'NoneType'
  File "/app/mylar/search.py", line 1906, in searchforissue
    foundNZB, prov = search_init(ComicName, IssueNumber, str(IssueYear), SeriesYear, Publisher, IssueDate, StoreDate, actissueid, AlternateSearch, UseFuzzy, ComicVersion, SARC=SARC, IssueArcID=IssueArcID, mode=mode, rsscheck=rsscheck, ComicID=ComicID, filesafe=Comicname_filesafe, allow_packs=allow_packs, oneoff=oneoff, manual=manual, torrentid_32p=TorrentID_32p, digitaldate=DigitalDate, booktype=booktype, ignore_booktype=ignore_booktype)
  File "/app/mylar/search.py", line 363, in search_init
    findit = NZB_SEARCH(ComicName, IssueNumber, ComicYear, SeriesYear, Publisher, IssueDate, StoreDate, searchprov, send_prov_count, IssDateFix, IssueID, UseFuzzy, newznab_host, ComicVersion=ComicVersion, SARC=SARC, IssueArcID=IssueArcID, RSS="no", ComicID=ComicID, issuetitle=issuetitle, unaltered_ComicName=unaltered_ComicName, allow_packs=allow_packs, oneoff=oneoff, cmloopit=cmloopit, manual=manual, torznab_host=torznab_host, torrentid_32p=torrentid_32p, digitaldate=digitaldate, booktype=booktype, chktpb=chktpb, ignore_booktype=ignore_booktype)
  File "/app/mylar/search.py", line 1308, in NZB_SEARCH
    elif all([int(F_ComicVersion) == int(findcomiciss), fndcomicversion is not None, booktype == 'TPB', filecomic['booktype'] == 'TPB', filecomic['justthedigits'] is None]):
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

PR #356, opened 2020-06-21
FIX: use regex against platform.system

barbequesauce:


barbequesauce:
Done, oops

HypKin:
hmm i'm still getting the same 500 error after switching to the dev branche via git and then updating via web gui

PR #358, opened 2020-06-22
Removed unrar-cffi component from ComicTagger and requirements.txt

evilhero:
unrar-cffi is now an optional component. Will work with either unrar-cffi if installed, or default back to included rarfile if it's not installed.

If it's already installed, Mylar3 will use it when unraring from within ComicTagger. Otherwise, will default back to the included rarfile module.

If using the rarfile module, either the unrar binary HAS to be in your environment PATH, or you need to locate the .ComicTagger directory path (either in the mylar root, or in the Mylar user's home directory). Once located, edit the ```settings``` file and set the ```rar_exe_path``` to the FULL PATH to the unrar binary (ie. /usr/local/bin/unrar).

Note: wiki updated to include the above with additional instructions as required [here](https://github.com/mylar3/mylar3/wiki/unrar-cffi-vs-included-rarfile)

### DIFF
--- a/lib/comictaggerlib/comicapi/comicarchive.py
+++ b/lib/comictaggerlib/comicapi/comicarchive.py
@@ -30,7 +30,13 @@ try:
 except ImportError:
     nsort = False
 
-from unrar.cffi import rarfile
+try:
+    from unrar.cffi import rarfile
+    unrar_cffi = True
+except ImportError:
+    from lib.rarfile import rarfile
+    unrar_cffi = False
+
 try:
     import Image
     pil_available = True
@@ -246,6 +252,8 @@ class RarArchiver:
     def __init__(self, path, rar_exe_path):
         self.path = path
         self.rar_exe_path = rar_exe_path
+        if unrar_cffi is False:
+            rarfile.UNRAR_TOOL = rar_exe_path
 
         if RarArchiver.devnull is None:
             RarArchiver.devnull = open(os.devnull, "w")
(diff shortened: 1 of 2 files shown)


evilhero:
```rar_exe_path``` gets populated to a None value on default. The rarfile attempts to use the unrar binary located in the path. If nothing gets hit then it errors I believe indicating as such (same as it did prior to Mylar3). Otherwise, once Mylar has started at least once, the .ComicTagger/settings file is generated and then the rar_exe_path can get populated manually at that point by the user if required.

professionaltart:
Just tested this.  
Win10 environment.  Was using python 3.7.7
Downloaded/installed 3.8.3
Upgrade pip, run -requirements
No break on unrar-cffi

I don't have any working Win10 VMs around atm to test with on a fresh environment, sorry.

Looking through the wiki page on it, I figured I'd just add the rar file to the config file so I don't forget it during SOE setup on next windows install.  Which made me realise the ct_settingspath in mylar's config.ini was set to an old homedir location that no longer exists. 🤦‍♀️

kanine:
The issues with ComicTagger (on Windows) seem to have come back again of late, I can't pin down exactly when the tagging started to fail again but I'm now seeing the following error:

`22-Feb-2023 16:06:36 - WARNING :: mylar.run.142 : CP Server Thread-8 : [META-TAGGER][WARNING] Make sure that you are using the comictagger included with Mylar.`

mylarr: v0.7.0
Python Version : 3.9.10
OS: Windows 10

I have configured the following in the settings file for comictagger:

`rar_exe_path = C:\Program Files\WinRAR\Rar.exe`

and the following in config.ini [Metatagging]

`ct_settingspath = D:\Apps\mylar3`

There don't appear to be any other clues in the log file.

evilhero:
@kanine please don't bump an almost 3yr old thread as we won't see it normally and you won't get a response. You should be creating a new github issue with your problems indicated.  

That being said, you need to point to the UnRaR binary, as that's a cli-only version.

If changing it to that doesn't help, please create a new issue with a carepackage with debug logging and we'll check it out.

Issue #321, opened 2020-05-28
One issue downloads then nothing until I restart Mylar?


paulr24:
Hi, I recently started using Mylar and I'm experiencing some strange issues. I have set up both NZB and DDL (GetComics) as search providers. For some reason upon opening Mylar it seems that it will often grab one issue of a comic via NZB, but then stop (but not always, sometimes it will not grab anything). If I close and re-open Mylar, it grabs one comic again, then stops. Via DDL comics seem to download at random as well, and it's rare.

What can I do to troubleshoot this? I have set logging to verbose and can post logs but not sure where to start since they're obviously long.

barbequesauce:
on the config page, click the "Care Package" button and post the zip here. It will give us system info, configuration data, a copy of your DB and log files....

paulr24:
[carepackage.zip](https://github.com/mylar3/mylar3/files/4699842/carepackage.zip)

paulr24:
> on the config page, click the "Care Package" button and post the zip here. It will give us system info, configuration data, a copy of your DB and log files....

Thanks, here you go!
[carepackage.zip](https://github.com/mylar3/mylar3/files/4699845/carepackage.zip)

paulr24:
I noticed this was marked invalid, is there anything more I need to do here for troubleshooting? Thanks!

barbequesauce:
No worries - invalid means "something aint right here". I'll take a look at your care package in a bit...

barbequesauce:
Looks like this exception is killing the search queue:

```
28-May-2020 03:08:16 - INFO    :: mylar.sender.42 : SEARCH-QUEUE : parameters set to {'apikey': '-REDACTED-', 'mode': 'addurl', 'name': 'http://192.168.1.-REDACTED-4:8090/api?apikey=-REDACTED-&cmd=downloadNZB&nzbname=Daredevil.013.2020.Digital.Zone-Empire.nzb', 'cmd': 'downloadNZB', 'nzbname': 'Daredevil.013.2020.Digital.Zone-Empire.nzb', 'output': 'json', 'priority': '-100', 'cat': 'comics'}
28-May-2020 03:08:16 - INFO    :: mylar.sender.43 : SEARCH-QUEUE : sending now to http://-REDACTED-.mynetgear.com:8080/api
28-May-2020 03:08:16 - ERROR   :: mylar.excepthook.315 : SEARCH-QUEUE : Uncaught exception: Traceback (most recent call last):
[...]

Issue #357, opened 2020-06-21
readlist - downloads via opds no longer auto marked as "read" is this a bug?


HypKin:
**Describe the bug**
I'm not sure if this is a bug or if this was a bug before.

i'm using chunky comic reader on ios and download comcis via opds. mostly via the readling list feature.

I havent read a lot in the last few months (as there wasnt a lot to read) but i started again in the last few days.

so i just found out, that comics that i download via the reading list opds feed are no longer marked as "read". this was i nice behaviour for me. is this how its supposed to be or did i stumble on a bug?

**CarePackage**
500 error as mentioned in my other issue

**Environment (please complete the following information):**
 - OS: windows 10
 - Master  e29114901810346e52486c8681c6b154639bcf86 (master)

barbequesauce:
can you either switch to the dev branch and update, so you get the new carepackage fix, and then attach a carepackage? Or attach a debug log please?

there were a few tweaks recently around opds that might have had some unintended side effects...

HypKin:
yep, i got you a debug log until then, but i don't see anything related to opds in the logs. 
[mylar.log](https://github.com/mylar3/mylar3/files/4809810/mylar.log)

barbequesauce:
OK - there's some additional logging in place that should help us trace what's happening. Can you switch to dev and update, then try to retrieve via OPDS please?

HypKin:
[mylar.log](https://github.com/mylar3/mylar3/files/4818519/mylar.log)
the log is bit long i had to work in between the update and restart. i started the opds download at about 11:22

barbequesauce:
Wow, that was staring me in the face.... #364 should fix this for you once commited.

barbequesauce:
Yeah no, I was barking up the wrong tree.

@evilhero from the log we clearly get to opds.py line [97](https://github.com/mylar3/mylar3/blob/c24e151d2d078cb3513a0b6b7dc7af02cf4b7a3e/mylar/opds.py#L97)... line 98 calls markasRead, but we never see readinglist.py line [115](https://github.com/mylar3/mylar3/blob/c24e151d2d078cb3513a0b6b7dc7af02cf4b7a3e/mylar/readinglist.py#L115) - not sure why? It's like the thread hangs...

barbequesauce:
@HypKin  that seems to have resolved it from my testing - can you give it a shot please?

HypKin:
[...]

PR #360, opened 2020-06-23
FIX: Nzbget param elimination

barbequesauce:
Helps with #321, doesnt fix tho


evilhero:
We already got #357 from the previous PR - can't merge cause it's not current. Gotta update and re-PR I think

PR #364, opened 2020-06-23
FIX: readlist instead of readinglist

barbequesauce:
Yo dawg, I heard you liked readinglists so I gotchu a readlist for your readinglist


barbequesauce:
Actually, the object name is [readlist](https://github.com/mylar3/mylar3/blob/d14353a24c0997cc095fd5206bf509625b760aef/mylar/opds.py#L671)...

barbequesauce:
Yeah I'm an idiot, nvm

PR #370, opened 2020-06-27
Mostly working streaming OPDS implementation

Micket:
Some bad news; trying to get this to work with android clients, i found out that
1. Kuboo hardcodes a whole bunch of ubooquity specific urls instead or using the links that the OPDS requests gives it. I can't get it to work. 
2. Challenger Viewer also adds a bunch of nondocumented variables to the url. Even trying to accomodate those, I couldn't get it to work.

So, neither of them follows the spec at all. :-1: 

Everything looks right in my manual inspections though.


barbequesauce:
Works beautifully with ChunkyTNG! Thank you and nice work!

Issue #400, opened 2020-07-22
Extra script fails to process: TypeError: can only concatenate str (not "bytes") to str


SenorSmartyPants:
My extra post processing scripts fail when called from mylar. They run fine from the commandline, so it is not a permissions issue.
[carepackage.zip](https://github.com/mylar3/mylar3/files/4961439/carepackage.zip)

I think it has to do with how the subprocess is opened.

https://github.com/mylar3/mylar3/blob/28d9812f1e38776f74bd26dc01264f205354318b/mylar/PostProcessor.py#L191

Add text=true to the arguments of Popen, this should set the script output to be string instead of bytes and solve this issue.
https://docs.python.org/3/library/subprocess.html#module-subprocess

evilhero:
This actually has zero to do with the Popen call itself - the problem is the logger line. It's trying to concat the result using the old ```+``` method, when it should be using the ```%``` method.

SenorSmartyPants:
#401 fixes this. My scripts run, with output displayed when debugging turned on in the log. 

Thanks.

PR #401, opened 2020-07-22
FIX: correction to popen call returning bytestream

barbequesauce:
Per #400 Thank you @SenorSmartyPants


SenorSmartyPants:
All Popen calls should be updated.

evilhero:
No need to update any Popen calls. This is a logger error, pure & simple. Modifying the line:
```    logger.fdebug("Script result: " + out)```
to:
```    logger.fdebug("Script result: %s" % out)``` 
would resolve it.

SenorSmartyPants:
I'm cool with whichever fix that gets implemented. Just want to write more post processing scripts and have them run. Thanks for all your work.

PR #414, opened 2020-08-16
fixes #413 also check if "/" is in size instead of only checking "//"

### LINKED ISSUE #413: Search for DCeased: Hope At World's End #007 returns an exception in ddl size calculation
**Describe the bug**
There seems to be a metadata quality issue with this getcomics package https://getcomics.info/other-comics/2020-07-29-weekly-pack/
If this package is returned in the searches the search crashes.
This seems to be because mylar expects "\\" to split the size of the package but this package only uses "\".
I think in getcomics.py after line 150 should be another if fork which handles this issue.

Issue can be reproduced with "DCeased: Hope At World's End #007"

Could you please take a look.
Thanks

Best regards
miracle152005

**CarePackage**
[mylar.log](https://github.com/mylar3/mylar3/files/5081140/mylar.log)

 - OS: docker image on ubuntu 18.04
 - Branch / commit:  ba72dcd0ea6aee2f4be5f634655cb204cd3842ac

miracle152005:
As descibed here the search aborts if only "/" is used to split sizes on getcomics:
https://github.com/mylar3/mylar3/issues/413

Please let me know if this fix works for you.

### DIFF
--- a/mylar/PostProcessor.py
+++ b/mylar/PostProcessor.py
@@ -188,7 +188,7 @@ class PostProcessor(object):
         logger.fdebug("Executing command " +str(script_cmd))
         logger.fdebug("Absolute path to script: " +script_cmd[0])
         try:
-            p = subprocess.Popen(script_cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, cwd=mylar.PROG_DIR)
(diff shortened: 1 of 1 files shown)


evilhero:
Cheers for the PR.

It's good to go, except for the merge PR in the list of commits. You're going to have to remove the extra commit ( update your branch prior to the PR creation ), and then resubmit it and I'll push the PR.

PR #474, opened 2020-10-01
Pass missing variables c_number and c_alpha to NZB_SEARCH

kalinon:
c_number and c_alpha are used on line 599 but never passed to the function.

### DIFF
--- a/mylar/search.py
+++ b/mylar/search.py
@@ -340,7 +340,7 @@ def search_init(ComicName, IssueNumber, ComicYear, SeriesYear, Publisher, IssueD
                     #if searchprov.lower() == 'ddl':
                     #    prov_count+=1
                     #    continue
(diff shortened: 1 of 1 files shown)


evilhero:
Not discounting this by any stretch here, as I can see the reference being used without a link. But you would only experience an error if you were searching for something that had a number and an alpha character as the issue number. Can you confirm the error you got for this one?

The only problem with passing everything from search_init is that other parts of Mylar call nzb_search directly, so if it was due to a filename search as per above this patch wouldn't fix the call from one of those stragglers.

Knowing how you encountered the error would help indicate if this patch would work on all around basis, or if it needs to be modified slightly.

Cheer for the PR work regardless ;)

kalinon:
I mainly noticed it via the PyCharm IDE code inspection. I have not ran into a specific error. I actually just spent a little time letting the IDE correct issues that were fairly straight forward. Im gonna open a PR shortly.

kalinon:
#476 is a more .... thorough... cleanup. Perhaps completely unnecessary. But the file is so large, I find its difficult to parse/contribute to it with all the warnings and unused vars hanging around.

evilhero:
Closing this as search.py has since gone thru both black + flake8. 

If there are still problems, we'll take it from a clean slate ;)

Thanks for bringing this to light!

PR #475, opened 2020-10-01
ensure that SEARCHLOCK is released

kalinon:
As far as i can tell `SEARCHLOCK` is only locked within `searchforissue`, however due to returns or possible errors, it can not be released. this simply wraps the function in a try/finally block which will release the lock regardless.


evilhero:
I have something similar in my local repo already done - except I included a seperate logging error so that the search errors can be referenced in the GUI.

I'll compare this one to mine and see how similar they are (if they are, I'll push yours and just merge in my logging additions in after the fact), as I can't recall if there were other fixes I did.

evilhero:
Closing this as search.py has since gone thru both black + flake8. 

If there are still problems, we'll take it from a clean slate ;)

Thanks for bringing this to light!

PR #477, opened 2020-10-02
feat: Added Signal sms notification

Aerex:
**Context**: Added a sms notification using Signal.

**Setting Up**:
1. Install [signald#v0.7.0](https://github.com/thefinn93/signald/tree/0.7.0) on your personal server
2. [Link](https://github.com/thefinn93/signald/tree/0.7.0#link) device to signald
3. Configure Mylar with your phone numbers (I use the same phone number for sending and receiving) 

signal_notif_mylar3
mylar_signal_error_not_found


barbequesauce (commented):
Looks good to me except requirements... if I'm not using Signal I shouldn't have to install signald. Of course I have no idea how to make it recommended in requirements.txt...

Aerex:
I think there is way to do that if you have a  setup.py but not so certain with requirements.txt. Let me look into it and get back to you. @barbequesauce

barbequesauce:
Any word @Aerex ?

Aerex:
Hey sorry about the late response. I was trying to find a seamless solution but couldnt find one. There are two workarounds that I can implement.

1. When the user enables Signal check system if the package is installed. If it is not notify the user via a message that the package needs to be installed. We can point the user to a page on the Githu Wikia for installation instructions
2. Dynamically install and uninstall package when Signalis enabled or disabled.

I would suggest workaround 1 since the alternative workaround is more of a security risk. Thoughts ? Thanks for keeping up on this PR. I think this would be a cool addition to mylar. @barbequesauce

barbequesauce:
Yep, I like option 1 as well. Option 2 is problematic in that (in most cases) the mylar user will not have the ability to add packages on the fly.

Side note, just thought of it as I was typing this - how does this impact windows users?

Aerex:
I will have to look into that as well. The Python package used in this PR requires you to have the [signal service](https://git.callpipe.com/finn/signald)
running on the same machine as mylar. The service can run on docker which is the recommended usecase for a windows user. 

I will do some testing on a windows machine to see how well mylar works inconjuction with the signals service under docker.

barbequesauce:
Looking good. I'll give this a spin next week on raspbian - let us know how it works out on windows please!

barbequesauce:
Umm you're going to have to resubmit this cleanly once you've worked out the kinks.... that second-to-last commit (merge python3-dev) will cause a loop in our commit history...

Aerex:
@barbequesauce Alright. I will recreate the PR against python3-dev

Issue #485, opened 2020-10-09
'zeroadd' referenced before assignment with Post-Processing


PCurd:
**Describe the bug**
After starting a new Mylar3 docker and migrating parts of the config file from Mylar 2 over (SAB config, etc.), when loading comic files from disk in a manual post-process an error is received for many filenames - error received is:

```
Uncaught exception: Traceback (most recent call last):
File "/app/mylar/logger.py", line 337, in new_run
old_run(*args, **kwargs)
File "/usr/lib/python3.8/threading.py", line 870, in run
self._target(*self._args, **self._kwargs)
File "/app/mylar/PostProcessor.py", line 1749, in Process
self.Process_next(comicid, issueid, issuenumOG, ml, stat)
File "/app/mylar/PostProcessor.py", line 2317, in Process_next
prettycomiss = str(zeroadd) + str(iss)
UnboundLocalError: local variable 'zeroadd' referenced before assignment
```
Full error:
```
2020-10-09 16:17:23 | ERROR | Uncaught exception: Traceback (most recent call last):File "/app/mylar/logger.py", line 337, in new_runold_run(*args, **kwargs)File "/usr/lib/python3.8/threading.py", line 870, in runself._target(*self._args, **self._kwargs)File "/app/mylar/PostProcessor.py", line 1749, in Processself.Process_next(comicid, issueid, issuenumOG, ml, stat)File "/app/mylar/PostProcessor.py", line 2317, in Process_nextprettycomiss = str(zeroadd) + str(iss)UnboundLocalError: local variable 'zeroadd' referenced before assignment
-- | -- | --
2020-10-09 16:17:23 | DEBUG | issue detected less than 10
2020-10-09 16:17:23 | DEBUG | [POST-PROCESSING] Zero Suppression set to : None
2020-10-09 16:17:23 | DEBUG | [POST-PROCESSING] issuenumOG: 2
2020-10-09 16:17:23 | DEBUG | [POST-PROCESSING] issueid: 797888
2020-10-09 16:17:23 | INFO | [POST-PROCESSING] [1/7] Starting Post-Processing for Fishkill issue: 2
2020-10-09 16:17:23 | INFO | [DUPECHECK] Duplication detection returned no hits. This is not a duplicate of anything that I have scanned in as of yet.
[...]

PR #486, opened 2020-10-09
FIX:(#485) If padding was not enabled, would still try to pad some values incorrectly

evilhero:


barbequesauce (approved):
Never realized padding was so complex

PR #490, opened 2020-10-11
Fixed js paths for webviewer

nymitr:


data/interfaces/default/header.html
```diff
@@ -8,8 +8,8 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
   <link rel="stylesheet" type="text/css" href="interfaces/default/css/materialize.min.css">
```
evilhero, line 9: This should point to the theme dir and not just the default one if we're correcting stuff (```href = interfaces/${interface}/css/materialize.min.css```)

evilhero (changes requested):
PR is done against master branch, not python3-dev. Please resubmit against the python3-dev branch as we don't merge PR's directly into master. Cheers

nymitr:
Sorry, I wasn't clear which branch should have been used.
Resubmitted.

data/interfaces/default/header.html
nymitr: Done.

evilhero (approved):

evilhero:
Good to go in development now. Thanks for the PR and the additional changes ;)

PR #493, opened 2020-10-16
fixed setting the tools.sessions.timeout value

claym:
moved 

`tools.sessions.timeout = options['login_timeout']`

to somewhere it would actually work :D


evilhero (approved):

evilhero:
Good to go - thanks @claym! I dunno wtf I was thinking when I did that initial reference.

 Nice catch!

PR #500, opened 2020-10-19
Fix the behinds

evilhero:
This 'should' correct the 44 behind problem and get everything back on track in regards to the git history due to some f'd up commits previously that were done to the master branch I guess.

This will still leave the 13 commits ahead, which is technically already in the python3-dev branch - but according to gitkraken, this PR lines everything back up in sync again. 

I need a gitkraken / git guru to look at this first please :)


barbequesauce (approved):
Confused AF but ok

Issue #513, opened 2020-10-23
Comic with "LR" in the issue number crashes post-processor


cwhits:
**Describe the bug**
It appears that an issue with "LR" in the issue number crashes the post processor. Interestingly, issues with "MU" in the issue number don't seem to have this problem.

**CarePackage**
[carepackage.zip](https://github.com/mylar3/mylar3/files/5430748/carepackage.zip)

The care package is attached but here's the relevant portion of the logs for search-ability (and so other users don't have to download the care package to see the error).
```
23-Oct-2020 12:40:02 - INFO    :: mylar.Process_next.2195 : Thread-13 : [POST-PROCESSING]  [1/1] Starting Post-Processing for The Amazing Spider-Man issue: 50.LR
23-Oct-2020 12:40:02 - DEBUG   :: mylar.Process_next.2198 : Thread-13 : [POST-PROCESSING] issueid: 813276
23-Oct-2020 12:40:02 - DEBUG   :: mylar.Process_next.2199 : Thread-13 : [POST-PROCESSING] issuenumOG: 50.LR
23-Oct-2020 12:40:02 - ERROR   :: mylar.excepthook.315 : Thread-13 : Uncaught exception: Traceback (most recent call last):
  File "/app/mylar/logger.py", line 337, in new_run
    old_run(*args, **kwargs)
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/app/mylar/PostProcessor.py", line 1749, in Process
    self.Process_next(comicid, issueid, issuenumOG, ml, stat)
  File "/app/mylar/PostProcessor.py", line 2254, in Process_next
    if int(iss_decval) == 0:
ValueError: invalid literal for int() with base 10: 'LR' 
```      

**Environment (please complete the following information):**
 - OS: Docker
 - Branch / commit: Version: v0.4.7, refs/pull/500/head (master)

**Additional information**
None

evilhero:
Yeah I fixed this everywhere else a few days ago, but forgot the post-processing module has a seperate hook.

Will look into in abit, not sure when it will hit master tho.

As a workaround, you can manually move the file into the series directory, do a Recheck Files and then a manual metatag on that issue if you want to add it to your collection.

barbequesauce:
good to close?

cwhits:
I'll leave that up to @evilhero

kingfenix:
[...]

Issue #505, opened 2020-10-20
Webviewer displays a large X in a circle when trying to view a comic


jwm-evans:
**Describe the bug**
Whenever I try to read a comic using the built-in webviewer, all I see is a large X in a circle (screenshot attached). This occurs for any comic that I try to view. Looking at the console log, it seems that every comic appears to have 0 pages, though other readers do not have any issues.
```
192.168.0.35 - - [20/Oct/2020:20:05:53] "GET /read_comic?ish_id=721067&page_num=0&size=high HTTP/1.1" 200 1252 "http://192.168.0.46:8090/comicDetails?ComicID=115725" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36"
192.168.0.35 - - [20/Oct/2020:20:05:53] "GET /interfaces/carbon/fonts/roboto/Roboto-Regular.woff2 HTTP/1.1" 404 467 "http://192.168.0.46:8090/interfaces/carbon/css/materialize.min.css" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36"
192.168.0.35 - - [20/Oct/2020:20:05:54] "GET /interfaces/carbon/fonts/roboto/Roboto-Regular.woff HTTP/1.1" 404 466 "http://192.168.0.46:8090/interfaces/carbon/css/materialize.min.css" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36"
192.168.0.35 - - [20/Oct/2020:20:05:54] "GET /interfaces/carbon/fonts/roboto/Roboto-Regular.ttf HTTP/1.1" 404 465 "http://192.168.0.46:8090/interfaces/carbon/css/materialize.min.css" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36"
[20/Oct/2020:20:05:54] HTTP 
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/cherrypy/_cprequest.py", line 638, in respond
    self._do_respond(path_info)
[...]

PR #517, opened 2020-10-26
Python3 dev merge

evilhero:
- FIX: log_level in config.ini will now honoured. If -v or -q is specified on execution, that will over-ride the config value

- IMP: Lessen some of the removal restrictions in carepackage so the logs are a bit more readable

- FIX:(#505) Fixes webviewer throwing a large X on new image load (just the cover)

- FIX:(#513) Fix for LR issues not being detected during post-processing runs


barbequesauce (approved):
"Here I come to save the day"

PR #518, opened 2020-10-27
IMP: Traceback/exception handling added to better handle crashes during search

evilhero:
- Try / Except handling around the main aspects of the search and weeklypull modules, which should make every traceback therein not queue-ending (meaning even after the error, the searches can continue as if they never happened).
- Added a special popup in the Logs tab (```History``` / ```View Logs``` / ```Exceptions / Tracebacks```) where all recorded tracebacks are logged - ability to view specifics, delete entries, and view log snippets pertaining to that specific traceback error, just prior to the error occuring.
- Fixed a problem with DDL provider being lower-cased which might have resulted in some missed searched calls.

- Formatted entire search.py file with black to conform to a soft-line limit of 88.
- linted search.py with flake8

### DIFF
--- a/data/interfaces/carbon/css/style.css
+++ b/data/interfaces/carbon/css/style.css
@@ -1783,6 +1783,23 @@ div#artistheader h2 a {
   min-width: 160px;
   text-align: center;
 }
+
+#exceptions_table th#date { text-align: center; max-width: 20px; }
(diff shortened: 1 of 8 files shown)


barbequesauce (approved):
search.py Jesus take the wheel

PR #529, opened 2020-11-01
fixes #528 SABNZB Path calculates mylar address incorrect

### LINKED ISSUE #528: Sabnzb is no longer able to download the packages
**Describe the bug**
Sabnzb is no longer able to download the packages.
The reason for this is that search.py line 4126, 4139, 4150, 4155 use HTTP_ROOT instead of the Port.

**Additional information**
I will provide a PR shortly to fix the issue

miracle152005:
SABNZB was no longer able to download because the address to get the nzb from was not correct.

### DIFF
--- a/mylar/search.py
+++ b/mylar/search.py
@@ -4123,7 +4123,7 @@ def searcher(
                     # if mylar's local, use the local IP already assigned to LOCAL_IP.
                     mylar_host = (
                         '%s%s:%s%s'
-                        % (proto, mylar.LOCAL_IP, mylar.CONFIG.HTTP_ROOT, hroot)
+                        % (proto, mylar.LOCAL_IP, mylar.CONFIG.HTTP_PORT, hroot)
                      )
                 else:
                     if mylar.EXT_IP is None:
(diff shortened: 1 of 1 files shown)


mylar/search.py
```diff
@@ -4147,12 +4147,12 @@ def searcher(
                             mylar_host = (
                                 '%s%s:%s%s'
                                 % (proto, mylar.CONFIG.HTTP_ROOT,
```
evilhero, line 4149: This should be ``` % (proto, mylar.CONFIG.HTTP_HOST, mylar.CONFIG.HTTP_PORT, hroot)```

evilhero (changes requested):
Stupid typos during flaking.

Just need that one correction that I mentioned, and I'll push this. 

Thanks for bringing it to our attention.

mylar/search.py
miracle152005: Did not see that one.
Should be fixed now

evilhero (approved):
good to go!

PR #533, opened 2020-11-04
Fix Mako error in opds.html

cnewham699:
Fix for a mako error when trying to use opds:

CompileException: Fragment 'for link in opds['links']: <% linktitle = '' if 'title' in link: linktitle =' is not a partial control statement in file '/volume1/mylar/share/mylar3/data/interfaces/default/opds.html' at line: 16 char: 1
 
fixed by reverting some of the original file formatting from a previous version of opds.html

### DIFF
--- a/data/interfaces/default/opds.html
+++ b/data/interfaces/default/opds.html
@@ -13,15 +13,20 @@
     <name>Mylar Server</name>
     <uri>https://github.com/mylar3/mylar3</uri>
   </author>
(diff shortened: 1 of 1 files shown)


evilhero (approved):

evilhero:
Nothing breaking in the PR - probably due to the opds file being black + flake8.

Thanks for the PR! :)

Issue #544, opened 2020-11-13
Manual Post-Processing does nothing and shows exception in logs


iammortimer:
**Describe the bug**
Doing a manual post-processing does nothing and shows an exception error in the logs

**CarePackage**
[carepackage.zip](https://github.com/mylar3/mylar3/files/5537992/carepackage.zip)

**Environment (please complete the following information):**
 - OS: Windows 10 2004 (19041.572)
 - Branch / commit: master [v0.4.6]
 - Python: 3.7.6

evilhero:
Yeah, there's no traceback / exception error in the logs that pertain to a manual post-process. Only thing is involving something regarding the story arcs

Unfortunately, your logs are at an INFO level, which means they are pretty much useless when it comes to trying to figure out what it is that you're having a problem with.

I suggest you update your version of Mylar (you're 3 versions behind), put Mylar into debug/verbose mode, and then try the manual post-processing again. If you still get an exception, then create a new care-package and post that.

Note that in v0.4.9 (the latest build), under History, View Logs, View Exceptions/Tracebacks you can see at a glance any exceptions that were triggered recently - you can also view a brief log snippet that might save pasting in a carepackage depending on what the issue presented is.

iammortimer:
Updated, but still results in the same exception (that isn't showing up in your exception screen btw)

[...]

PR #545, opened 2020-11-14
Update helpers.py

iammortimer:
Fix for #544


barbequesauce:
Makes sense to me - evilhero ?

evilhero:
Should be ok? The only concern I have is the 'invalid' aspect - I'm not sure if that's the best course of action to default the filenames to, as that could possibly throw exceptions elsewhere as the function is used everywhere.

I'll run thru a few tests on my side to see, but aside from that aspect it would be fine otherwise.

evilhero:
Awaiting response to #544 in order to finalise solution.

barbequesauce (approved):

Issue #257, opened 2020-04-20
Method for mass-updating database


Serneum:
**Describe the request**
This may be a bit niche, but this is the second time I've had issues with the drive that Mylar uses for comic storage where I've lost my entire comic collection (3k or so comics). Last time it happened, I went into the UI to update my comics from Downloaded, and the rescan put them into an Archived state. I then tried to update them to Wanted, but was only able to do them in batches of 25 without errors. Anything more than that and the application wouldn't update any of the comics at all. I ended up writing a script to hit some endpoints in the application to force the update myself, but a quick way to update my DB would be nice, even if it's just some instructions on how to connect to the DB and run an `UPDATE` query against all of the issues.

An API to update comics would also work so that I do not need to monitor the Network tab in my browser to hack together URLs that can accomplish what I need to do.

**How did you come up with the idea?**
Catastrophic hard drive failure, twice very close together.

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Branch / commit: Not sure at the moment, but I know it was up-to-date as of two weeks ago. My server is currently down until I get new storage drives in

**Additional information**
I understand that the idea is very specific to my needs, possibly to the point where I am the only person that ever uses it, and may not be worth any of your effort.

mylar3:
Yeah the typical rescan will check to see if the file exists where it used to be according to the db in order to flag it as ```Downloaded```. Because the files were no longer present, the recheck assumed they were manually moved off / deleted and marked them as ```Archived```.

Have you tried going into the ```Manage``` tab \ Manage Issues and then selecting ```Downloaded``` from the dropdown to show all the issues that are in a Downloaded status? You can then set the marker to show All the items, and use the other dropdown to set the status for the checkmarked items to Wanted.

As far as the API, I'm assuming this is akin to the option to ```Mark all skipped issues as wanted``` - save for the fact that you want to just force them all to Wanted. Definitely doable, the API is in a vast need for an update regardless - but if you would be able to indicate what it is that you want the update to hit on (update the status, recheck, etc) we can go from there.

Serneum:
Yeah, I've used the Manage tab with limited success. Anything over 25 records seems to throw an error, if I recall correctly. With 3000 comics to update, it takes a long time

As for the API, updating the status would be ideal. Also forcing the recheck that way would work. Currently, I can get by and change everything that is Downloaded to Wanted, but I could also rescan and then change everything that is Archived to Wanted.

Serneum:
[...]

PR #556, opened 2020-11-25
IMP:(#257) added changeStatus and recheckFiles to API

evilhero:
- IMP: API method **changeStatus** added to allow for mass changing of status' from a specific status to another, by comicid:
        - change **status_from** of every issue in series to specified **status_to**
        - if no comicid specified will mark ALL issues in EVERY series from status_from to specific status_to
        - required fields: **status_to**, **status_from**. Optional: **id**  (which is the ComicID if applicable)

- IMP: API method **recheckFiles** added to allow for batch rechecking of series based on comicid:
        - allow either individual / bulk recheck Files based on ComiciD
        - multiples are allowed as long as in a list: {'id': ['100101', '101010', '20181', '47101']}

### DIFF
--- a/mylar/api.py
+++ b/mylar/api.py
@@ -1,5 +1,4 @@
-# -*- coding: utf-8 -*-
-
+#-# -*- coding: utf-8 -*-
 #  This file is part of Mylar.
 #
 #  Mylar is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


barbequesauce (approved):
Amazingly readable ;)

PR #557, opened 2020-11-29
Added encoding filenames for webviewer

nymitr:
If there is not-encoded `#` symbol in filename, browser won't open file.


barbequesauce (approved):
Thank you for the contribution!

PR #558, opened 2020-11-29
Added next/previous page check

nymitr:
Avoid attempts to load next/previous page if there is no next/previous page.


barbequesauce (approved):
Nice! This was on the todo list, thanks!

Issue #553, opened 2020-11-24
OPDS Internal Server Error when fetching MetaInfo


jbecho:
I pulled today's update and tested the OPDS fix.  I no longer get the Mako Runtime Error but now have a new error.  When getting down to a comic to view the issues it is getting an internal server error.  This only occurs when fetching MetaInfo

'''
500 Internal Server Error
The server encountered an unexpected condition which prevented it from fulfilling the request.

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/cherrypy/_cprequest.py", line 638, in respond
    self._do_respond(path_info)
  File "/usr/lib/python3.8/site-packages/cherrypy/_cprequest.py", line 697, in _do_respond
    response.body = self.handler()
  File "/usr/lib/python3.8/site-packages/cherrypy/lib/encoding.py", line 219, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/cherrypy/_cpdispatch.py", line 54, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/app/mylar3/mylar/webserve.py", line 5882, in opds
    data = op.fetchData()
  File "/app/mylar3/mylar/opds.py", line 87, in fetchData
    result = methodToCall(**self.kwargs)
  File "/app/mylar3/mylar/opds.py", line 404, in _Comic
    issuedetails = mylar.helpers.IssueDetails(issue['fileloc']).get('metadata', None)
KeyError: 'fileloc'
'''

**Environment (please complete the following information):**
 - OS: Docker (https://hub.docker.com/r/linuxserver/mylar3)

evilhero:
Should be fixed in #560 (I think). 

When this hits the master branch you can verify the if the fix works at that point - should be a dev-to-master merge coming in the next few days.

barbequesauce:
yeah, I always wondered why I couldnt get that working.

PR #560, opened 2020-12-02
FIX:(#553) incorrect variable used to reference file location

evilhero:


barbequesauce (approved):
So that's why I could never get it to display anything

PR #571, opened 2020-12-15
CDH improvements & enhancements

evilhero:
- IMP: Added pause-detector (and readd to queue) to SABnzbd CDH handling. Will now handle if SABnzbd is in paused mode during a send. When SABnzbd is unpaused, will auto-requeue all items that were sent to SABnzbd via Mylar so that they can be automatically post-processed sequentially (it won't redownload them again if SABnzbd aleady downloaded them)
 
- IMP: Added additional CDH handlers to CDH for SABnzbd (queue_paused, failed_in_sab, unhandled status) for more descriptive handling of errors

- FIX: If incorrect volume was parsed previously from CV, would never update to correct volume unless manually entered.

- FIX: duplicate checker would fail during post-processing if item was not in the location specified and killed the queue.


mylar/sabnzbd.py
```diff
@@ -164,14 +175,17 @@ def historycheck(self, nzbinfo):
...
+                        break
                 elif hq['nzo_id'] == sendresponse:
                     nzo_exists = True
                     logger.fdebug('nzo_id: %s found while processing queue in an unhandled status: %s' % (hq['nzo_id'], hq['status']))
-                    break
+                    return {'failed': False, 'status': 'unhandled status of: %s' %( hq['status'])}
 
             if not nzo_exists:
-                logger.info("Cannot find nzb %s in the queue.  Was it removed?" % sendresponse)
+                logger.info('Cannot find nzb %s in the queue.  Was it removed?' % sendresponse)
```
barbequesauce, line 188: single quotes?

barbequesauce (approved):

PR #589, opened 2021-01-01
Add a config option for backups during startup

Serneum:
This should add the ability to run backups at startup through a config option instead of just a command line option. This will allow people like myself to use the LinuxServer.io image and maintain DB backups. This can be accomplished by restarting the container on a schedule, which is not ideal but is doable.

I'm getting errors when using the Dockerfile, so I have not been able to test this change.
```
ERROR: unsatisfiable constraints:
  git-2.24.3-r0:
    breaks: world[git=2.24.1-r0]
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,13 @@ FROM python:${BASE_VERSION}
 
 # set version label
 ARG MYLAR_COMMIT=v0.3.0
+ARG ORG=mylar3
 LABEL version ${BASE_VERSION}_${MYLAR_COMMIT}
 
 RUN \
 echo "**** install system packages ****" && \
  apk add --no-cache \
- git=2.24.1-r0 \
+ git=2.24.3-r0 \
  # cfscrape dependecies
  nodejs=12.15.0-r1 \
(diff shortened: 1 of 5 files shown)


Serneum:
I updated the Dockerfile to use Git 2.24.3-r0 and was able to build an image but unable to get it to start up:
```
  File "/app/mylar/Mylar.py", line 26, in <module>
    import mylar
  File "/app/mylar/mylar/__init__.py", line 38, in <module>
    from mylar import logger, versioncheckit, rsscheckit, searchit, weeklypullit, PostProcessor, updater, helpers
  File "/app/mylar/mylar/rsscheckit.py", line 21, in <module>
    from mylar import logger, rsscheck, helpers, auth32p
  File "/app/mylar/mylar/rsscheck.py", line 52, in <module>
    feedparser._FeedParserMixin._start_newznab_attr = _start_newznab_attr
AttributeError: module 'feedparser' has no attribute '_FeedParserMixin'
```

Any guidance would be appreciated

Serneum:
Ok, looks like this works for me locally now

barbequesauce (approved):

PR #592, opened 2021-01-02
Pull-list patch for year flipping to 2021

evilhero:
- FIX: fixed pull-list attempting to poll week 0, 2020 instead of week 52, 2020 or week 1, 2021 (resulted in empty pull-list)
- FIX: pull_refresh value in ini will now be written out after each pull refresh (it stores it in the db, but w/e)


barbequesauce (approved):
Deja vu all over again

Issue #464, opened 2020-09-23
Headers have lots of issues


BRTPOB:
The headers on the main display for Mylar in carbon don't line up properly, don't fit within their proper fields, and aren't centered properly. Screenshot attached.
![image](https://user-images.githubusercontent.com/7201028/94069827-25b44580-fde9-11ea-9a82-9fc4e55a92d4.png)

barbequesauce:
What browser? NAI in Safari / clean install / incognito....

<img width="999" alt="Screen Shot 2020-10-12 at 9 42 14 AM" src="https://user-images.githubusercontent.com/25070944/95753246-57ddf800-0c6f-11eb-9b1b-2f763b663e4c.png">

BRTPOB:
First off, it's NAB or CNR not NAI. Secondly, Firefox.

BRTPOB:
Addendum: initially it was fine on Chromium, but as soon as I went to This Week and clicked "Recreate Pull List", the tables redrew and I clicked the link for "Back to overview" and that's when the brokenness showed up.

barbequesauce:
Hush Az.

Yeah, still can't recreate in Chrome or Safari. There is a quick flash as the table is rendering where I see Last Issue as two lines, but when jquery finishes its thing it goes as above.

Any javascript console errors?

miracle152005:
I can confirm the issue exists.
It seems to be a Firefox specific issue and cannot be reproduced in chrome/edge.
Deactivating the font overrides for "table.display thead th" (font-weight and font-size) using the Inspector in Firefix results in the text matching the columns.
This seems to indicate different possibilies I can think of:
a) Firefox renders font sizes differently
b) Firefox applies a different order when applying the scripts.

But since the actual sizes for the column seem to match in Chrome and Firefox b seems unlikely.

miracle152005:
After switching back to default theme it also happens in the default theme now when using firefox.

barbequesauce:
#595 should address this...

BRTPOB:
Yay. Once I get a working install again, I'll check it out. Thanks dude.

PR #595, opened 2021-01-03
FIX: #464 table headers muffintop over adjacent columns

barbequesauce:
FIX: #464 table headers muffintop over adjacent columns


barbequesauce:
solution found https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout - use the demo table to see the differences

evilhero (approved):

Issue #386, opened 2020-07-03
Vertical alphaindex


barbequesauce:
as requested by our favorite QA engineer @BRTPOB

barbequesauce:
Still working on this...

barbequesauce:
evilhero figured it out!

PR #601, opened 2021-01-09
IMP: add ability to move alphaindex from the top to the left or right…

barbequesauce:
IMP: add ability to move alphaindex from the top to the left or right sides (or disable entirely) all from the config page

Note that the left or right positions do not function as advertised yet - **work in progress** - please do not merge yet.

If we have any data tables / css experts around, please look at lines 138 / 166 / 194 of index-alphaindex.html. The goal is what's seen [here](https://jsfiddle.net/o10wpcjv/1/); I suspect it's the use of three divs to build the table and controls are what's throwing it off. Current result appears like this when setting the position to the right:

Screen Shot 2021-01-09 at 6 45 11 PM


barbequesauce:
Note this fixes #386 for our friend @brtpob

barbequesauce:
@sukhysall or @ellnic any chance you could spare a few cycles to take a look at this? I'm fairly confident it's all css issues

ellnic:
@barbequesauce I only understand some of this, some of it is over my head. You might want to ask @ronilaukkarinen.

ronilaukkarinen:
Thanks for pinging @ellnic, however I myself don't use mylar, only [watcher](https://github.com/barbequesauce/Watcher3). Took a quick look at that CSS part and there's float and width used. They often cause problems in cases like this, it should be built with alternative ways like flex or grid. But without proper context I have no idea and no time right now to review this further, sorry guys!

barbequesauce:
Thanks for looking!

PR #602, opened 2021-01-10
FIX: "RSS Feeds" Job Stuck Running

davalanche:
My "RSS Feeds" Job, as seen on Manage -> Activity / Jobs (Schedulers), always appears to be Running after the initial / successful run of the job after startup.  Looking at the code, the "[RSS-FEEDS] Watchlist Check for new Releases" (mylar.search.searchforissue(rsscheck='yes')) call in rsscheckit.py does not reset mylar.SEARCHLOCK which causes the subsequent job to do nothing but sleep.


barbequesauce (approved):
Good stuff, thank you ! Another hole plugged

PR #605, opened 2021-01-13
A couple SABnzbd fixes

davalanche:


evilhero:
Yeah this will break compatibility with anything < SABnzbd 3. 

From the looks of it, they changed the apikey field from ```apikey``` to ```apikey_display```, so really all that should be needed is to check the version and search for the required keyname. I just don't know if keeping the ```find_apikey``` will work on < SAB3 instances, as they don't distinguish the apikey field names the same (meaning, we might have to keep the existing structure for non 3.0 installs but segment it based off of versions).

 I don't have a SAB < 3 installed - wonder if we can test this PR out prior (maybe @barbequesauce has an older version still).

evilhero:
Also,  the change to webserve to remove the ```Get API``` call was meant as a safety measure - if the user didn't click on the Get API button first (cause really does that suprise you), when it goes to test the connection, the api key field would be blank. So the safety measure was to run a quick check to get the key during a test connection when the api key wasn't present or invalid in some cases.

Again, the usefulness of the process is questionable - but at the time there was a need to have some catch for those users who either didn't put in an api key first and hit the Test Connection button, or entered it in wrongly in the first place. Whether or not this entire process is even used tho - I haven't a clue as it was years ago this was put in and just left as is for the most part (as I'm sure is obvious in some parts).

davalanche:
Thanks for the detailed feedback @evilhero, it is appreciated. :)

It looks like there are still SABnzbd 2.3.9 docker images available so I can pull one down and test against 2.x and then depending on the results rework the code to support both 2.x and 3.x .

evilhero:
With #606 basically hammering the changes in this PR - can you close this PR and submit a new PR with the changes that weren't in #606 - and then we can merge #606 and those new changes.

davalanche:
New Pull Request created under #615.

PR #606, opened 2021-01-14
FIX: SABnzbd "Get API" link not working (tested against SABnzbd 3.1.1…

davalanche:
… and 2.3.9)

This commit is intended to replace commit f1bb5b8 of PR #605.

Change verified to be working against SABnzbd 3.1.1 and 2.3.9.


barbequesauce (approved):
Thanks for the contribution!

PR #622, opened 2021-01-30
Made webviewer use correct relative path

nymitr:
Some cbr/cbz contain actual files in subfolder.


barbequesauce (approved):
Good catch, thanks for contributing !

PR #629, opened 2021-02-04
FIX: fixed sorting of Have column on non-alpha index

evilhero:
This PR only affects the view for the non-alpha main page (alphaindex = False)

Previously sorting the Have column on the main index page would result in sorting by completion percentage, which would incorrectly sort.

This fixes it so that it now sorts by Total Issues descending, then by Have Issues ascending - which results in items in a more logical sequence (ie. so 0/400 comes before 0/1, both of which come before 1/34).


barbequesauce (approved):
Guess I gotta replicate this in the AI page

PR #635, opened 2021-02-11
IMP: Renamed blacklisted_publishers to ignored_publishers

evilhero:
- Renamed blacklisted_publishers variable to ignored_publishers. Will auto-convert any existing values on startup.
- Updated how Mylar interprets the bad_definitions settings so it doesn't create extra commas as values when removing/replacing old variables.


barbequesauce (approved):
Good stuff

Issue #620, opened 2021-01-22
Opensuse Tumbleweed fails to pip3 the requirements


mgaulton:
I just found this repo and am trying to install it on my system to prep for migration.
I have removed the old install and put this in the same dir, all contents removed.
after a git clone I attempt the pip3 step and get the below. multiple attempts have failed.

`server6:/opt/mylar/mylar-master # pip3 install --no-cache-dir -U -r requirements.txt
Requirement already satisfied: beautifulsoup4>=4.8.2 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 8)) (4.9.3)
Requirement already satisfied: cfscrape>=2.0.8 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 9)) (2.1.1)
Requirement already satisfied: CherryPy>=18.5.0 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 11)) (18.6.0)
Requirement already satisfied: feedparser>=5.2.1 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 13)) (6.0.2)
Requirement already satisfied: Mako>=1.1.0 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 14)) (1.1.4)
Requirement already satisfied: natsort>=3.5.2 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 15)) (7.1.0)
Requirement already satisfied: portend>=2.6 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 17)) (2.7.0)
Requirement already satisfied: pytz>=2019.3 in /usr/lib/python3.8/site-packages (from -r requirements.txt (line 19)) (2020.5)
Requirement already satisfied: rarfile>=4.0 in /usr/local/lib/python3.8/site-packages (from -r requirements.txt (line 20)) (4.0)
Requirement already satisfied: requests>=2.22.0 in /usr/lib/python3.8/site-packages (from -r requirements.txt (line 21)) (2.25.1)
Requirement already satisfied: simplejson>=3.17.0 in /usr/local/lib64/python3.8/site-packages (from -r requirements.txt (line 22)) (3.17.2)
Requirement already satisfied: six>=1.13.0 in /usr/lib/python3.8/site-packages (from -r requirements.txt (line 23)) (1.15.0)
[...]

PR #651, opened 2021-02-23
FIX: Remove dependency on pyinstaller (#620)

barbequesauce:


barbequesauce:
Finally built a test rig and it seems to work without it.

evilhero (approved):

PR #658, opened 2021-03-01
IMP: When the weekly pull job runs, it will now recreate the pull for previous week as well

evilhero:
Due to CV not updating items until the following week in some cases, and because the weekly pull changes to the new week on Sundays, some issues on the Wanted list from the previous week would not get updated until the given series was either manually refreshed or a new issue came out in subsequent weeks.

This PR tries to ensure that the stragglers (ie. updates on CV that happen after the pull changes over), are updated and displayed correctly on the given week's pull. This will update the current week, as well as the previous week's pull-list with any relevant issue information changes.


barbequesauce (approved):
Yay for monkey patches!

PR #666, opened 2021-03-12
IMP: Added 'Add it all.' option to the Weekly page

evilhero:
```add it all.``` option  added to the Weekly page, which will add all series that are not currently on the watchlist into the user's watchlist for the currently viewed week.

This is a threaded call - so that if people are doing the ```add it all.``` option several times in succession for differing weeks, it will queue them all up in succession (FIFO). It also checks the queue list to see if a comicid is possibly queued to be added multiple times due to how issues are published, in which case it won't add it to the queue if it's already on the list.

This has been lightly tested - about 30 adds with multiple weeks being added in succession. No problems encountered, but I'm sure others will be able to stress-load it better than I.


barbequesauce (approved):
Fark you, Frank.

virtualadept:
> Fark you, Frank.

<3 Thanks guys, my hoarding has increased by 43.6%

PR #669, opened 2021-03-17
FIX: Correct padding 0x when issue number is between 10 and 100

### LINKED ISSUE #668: Number padding adds a 0 even if the number already has the correct number of digits
**Describe the bug**
When we choose a number padding, we expect the issue numbering to contains the exact number of digits.

For example, if we choose `0x` we expect issues 0-9 to be 00-09 and issue 10 and over to keep the same numbering. Instead issues 0-9 got correctly renamed to 00-09, but issues 10+ got renamed 010, ... instead of keeping the same number.

 - OS: Linux / Docker
 - Branch / commit: Master

foux:
Do not add a 0 for issues between 10 and 100 when padding settings is `0x`. Fixes #668


evilhero:
Does not work for numbering above 100, and displaces what the entire padding process is supposed to do.

This will get addressed hopefully when a future PR gets pushed which will provide more padding options for users.

PR #675, opened 2021-03-24
FIX: fix for image notifs and issue image display via popup not displaying cbz files that contained .jpeg images

evilhero:


barbequesauce (approved):
srsly

Issue #277, opened 2020-05-03
Hide Ended & Have All


ElementSixtySix:
Possibility to filter series that have ended and where all issues are in collection.

Allows user to more easily see ongoing and/or incomplete series.

barbequesauce:
I have a POC working for the series page that filters out continuing/ended, along the lines of how we do wanted/archived/downloaded/snatched etc on the comic details page. I think this addresses your first case.

<img width="344" alt="Screen Shot 2020-07-06 at 7 47 08 PM" src="https://user-images.githubusercontent.com/25070944/86669336-7f169080-bfc1-11ea-9520-15668fe208a0.png">


The second case is a bit more fuzzy - you can do this with the "filter comic" box above the pagination controls; were you lookibg for something more specific? I'd hesitate to add more complexity to a busy screen already...

barbequesauce:
Better late than never... #695 finishes out the pic above, should be merged to dev sometime this weekend.

barbequesauce:
Closing due to staleness, we welcome contributions for new features. If this is an issue and you are still experiencing it, please open a new ticket with a care package attached.

PR #695, opened 2021-04-03
IMP: Display/hide continuing or ended series on the front page (#277)

barbequesauce:


evilhero:
We're reverting changes in helpers.py from a PR the other day. Don't see any problems other than that. 

I think you'll have to rebase clean, cause if you commit adding them back, it'll conflict when we go to merge dev - - master.

barbequesauce:
Aghhhhhhh ok sorry

PR #696, opened 2021-04-03
IMP: Add filtering checkboxes for continuing & ended series (#277)

barbequesauce:


evilhero:
I think you might have forgotten the javascript stuff - the checkboxes aren't doing anything when I try them (they toggle on / off, but they don't change the table at all). 

I tried both the default and carbon interface - didn't change the buttons being able to do anything unfortunately I'm guessing you modified the issCheck from the comicdetails, which would mean that a hidden select box would be in there too somewhere. I tried to add some stuff in, but I didn't know where you were pulling everything from, or if you just missed them due it being a CLI PR.

barbequesauce:
Will update and reopen (eventually)

Issue #652, opened 2021-02-24
Make symlinks relative for StoryArcs


Shalar:
I would like to request symlinks be relative (or the option given to be) rather than absolute. I run mylar3 in a docker container with my comics mounted as /comics, but the actual path is something else. If I try to use a terminal or sftp client to access the symlinks it fails due to the actual path not existing.

davalanche:
I can take a look at this request.

evilhero:
@davalanche , sure feel free to - this probably would have been a low priority for us given the current workload, so if you want to take a look at it, much appreciated!

evilhero:
Assuming fixed since this been in for a while now (and it's stale otherwise).

PR #697, opened 2021-04-05
IMP:(#652) StoryArcs: Added the option to make symlinks relative

davalanche:


evilhero:
Looks good from a coding aspect - much thanks for doing that/this.

One slight problem I have - the symlink option probably should only show as an option if the option is selected/enabled via the dropdown. Either that, or make the option as part of the drop down , just clarifying relative vs. absolute as a symlink option perhaps. Having the symlink option appear otherwise will just cause some confusion issues (ie. "I checked it off under directory, and selected copy but how come everything is being symlinked?" or vice-versa kinda deal).

If you're unsure about the enabling via dropdown, I think I can do that at some point - but adding the option to the dropdown is relatively simple, just would need to clean up the variable naming in the PR.

Cheers,

davalanche:
Thanks for the feedback @evilhero.

> symlink option probably should only show as an option if the option is selected/enabled via the dropdown.

Unless I am misunderstanding, this should already be happening via storyarc.html.  If softlink is selected the relative option appears and if anything else is selected the relative option is hidden.  If the relative option remains selected when say the copy action is selected it won't matter as relative/absolute is only relevant when action_op = softlink (file_ops(); helpers.py).

If you would prefer to see this option in the dropdown though I will refactor and submit another PR.

evilhero:
So in looking at this (cause the first comment I wasn't testing it, was just looking at the code) - it all worked just fine and was hidden when not selected via the dropdown.

However, the spacing was bugging me - both from the 'slid-up' checkbox, and the spacing to the bottom - so I made some slight adjustments (I only adjusted the html, nothing else):
- noticed that there was an extra ```</div>``` reference, so removed that.
- changed the ```(#arc_fileops option:checked").val()``` items to ```document.getElementById("arc_fileops").value``` since they're not checkboxes.
- removed the ```<div id="row checkbox left">``` since it was messing up the spacing.
- forced the styling of the ```<small>``` so that it would appear smaller with the same greyscale as the other items on the page.

These adjustments can be found in the ```davalanche-python3-dev``` branch here. If you can check it out and see if it's as to what you're expecting, you can either mod your existing PR or we can just merge from that branch if it's alright (your PR is in there regardless).

davalanche:
> These adjustments can be found in the davalanche-python3-dev branch in the mylar3 repo. If you can check it out and see if it's as to what you're expecting, you can either mod your existing PR or we can just merge from that branch if it's alright (your PR is in there regardless).

I reviewed the diff and pulled down storyarc.html and the formatting does look a lot better.  I think I spent more time trying to get the html/css halfway decent than I did on the actual python code.  :) I am OK with you merging in davalanche-python3-dev as it accurately reflects both of our changes.

The jQuery :checked selector works on the options of select elements as well as check boxes but agreed that given the context getElementById().value (or even option:selected) makes more sense.

Thanks!

evilhero:
Closing in favour of using modified #698.

PR #711, opened 2021-04-13
Bugfix: type() is overwritten by string

Eloque:
Currently, in webserv.py the type() function is being override as there is a string called "type"

This leads to all type comparisons failing. This is no issue when the comic is not already in the db. But when it is, it cause queries to fail as the statement `"haveit":      imp_cid,` when ready having a comic in the library is set to a dict in stead of string. 

The quick fix for this is to use 
`import builtins` and `builtins.type(x)' instead of 'type(x)`

I've done this in webserve.py, other files might have similar issues.

### DIFF
--- a/mylar/webserve.py
+++ b/mylar/webserve.py
@@ -27,6 +27,9 @@ import json
 import copy
 import ntpath
 
+# Because Type has been overwritten.
+import builtins
+
 from mako.template import Template
 from mako.lookup import TemplateLookup
 from mako import exceptions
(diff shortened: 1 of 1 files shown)


evilhero:
While it's appreciated that the PR has been brought up, it's probably better practice to rename the type instances that are being declared as a variable to something other than type, as opposed to specifically calling the type function itself.

I've done this for a few modules already, but haven't done every instance as of yet, so obviously a little behind. 

Also, for future reference, we only accept PR's that are done against the python3-dev branch, not the master branch.

Eloque:
> While it's appreciated that the PR has been brought up, it's probably better practice to rename the type instances that are being declared as a variable to something other than type, as opposed to specifically calling the type function itself.
> 
> I've done this for a few modules already, but haven't done every instance as of yet, so obviously a little behind.
> 
> Also, for future reference, we only accept PR's that are done against the python3-dev branch, not the master branch.

I agree on the principle of better practice, but didn't want to change the variable names just so. 
Shall I redo a PR on the correct branch using the suggested solution?

PR #718, opened 2021-04-16
fixed two api calls that were broken

ajurna:
These endpoints were using old non existent code. I've updated them to use the current method.


evilhero:
Looks good - just against the wrong branch ;)

We only merge PR's into python3-dev,  not directly into master. 

So you'll have to do a new PR against the dev branch and then we can merge it.

Issue #715, opened 2021-04-14
Suzuka (Del Rey) cannot be added


miracle152005:
**Describe the bug**
Suzuka issued by Del Rey cannot be added (https://comicvine.gamespot.com/suzuka/4050-40482/)
Most likely because of the last issue which is defined as Issue #13-15

Could you please take a look.
Thanks
Best regards
miracle152005

**CarePackage**
[mylar.log](https://github.com/mylar3/mylar3/files/6308408/mylar.log)

**Environment (please complete the following information):**
 - OS: ubuntu 18.04
 - Branch / commit:  ab381ca1113bbc1e775a868ecf2a2a9ac52744da

PR #726, opened 2021-04-20
Importing multi-issue Issues.

ajurna:
… raised with #715.

i have put in an algorithm to generate consistant issue numbers from issue spreads based on what was there.

There is still an issue where the issue number generation code is duplicated between both the import code and the helper code.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -1119,7 +1119,14 @@ def issuedigits(issnum):
                             logger.fdebug('this does not have an issue # that I can parse properly.')
                             return 999999999999999
                     else:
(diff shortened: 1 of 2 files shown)


ajurna:
i have another commit where i remove all the duplicated code and just use the helper function. but i don't know if there is a reason to calculate this data separately

evilhero (approved):

evilhero:
Looks good - and merged!

Will have to look at the helpers function you did before clearing it obviously - that function is called in a ton of places in order to get the integer value for calculations. I don't have a problem with trimming the duplicates, but will have to see what you have done thus far.

Cheers

PR #733, opened 2021-04-24
FIX: carbon interface search results row colouring

evilhero:
- FIX: (carbon) When viewing search results items already in library would have differing coloured bars, simplified down to one color instead of alternating
- FIX:(carbon) Search results print editions green colouring not so neon green


barbequesauce (approved):
But I like neon lime

PR #746, opened 2021-04-30
Updated readme.md to work with wiki linking

evilhero:
- Updated readme.md so that only bullet points are on the readme linking to related wiki posts.
- Tried to keep it as 'clean' as possible, and used sonarr's readme as the basis for the changeup ( much <3 to the guys )


barbequesauce (approved):
Nice!

Issue #846, opened 2021-08-04
Manage Issues Page - Possible Memory leak when managing SNATCHED issues.


royaljack:
**Describe the bug**
When attempting to manage SNATCHED issues via  http://MyIP:8090/manageIssues?status=Snatched my memory usage jumps from about 250MB of usage to over 30GB of RAM and all 7 cores Mylar3 is assigned to peg out at 100%. After about 20 minutes mylar's docker crashes/restarts. 

**CarePackage**
On the config page, there is a button labelled "CarePackage". Pressing this button will download a zip file containing a redacted copy of your config, your database, a summary of your operating environment, and your logs. Upload that file here.
[carepackage.zip](https://github.com/mylar3/mylar3/files/6928258/carepackage.zip)

**Environment (please complete the following information):**
 - OS: Unraid 6.92 Docker 20.10.5 Python 3.8.10
 - Branch / commit: v0.5.6 ( master)

**Additional information**
I do have quite a large watch list and many series. Day to day operations work fine and new issues are snatched and postprocessed successfully each week. The only time I have any problems is specifically when I try to mass manage snatched issues.

evilhero:
PR #850 will fix the memory leak on the Manage Issues page (it will still be slow to load, but it won't crash out or jump the memory).

Future update will have the page load dynamically instead of having to load the entire dataset before it can render the page.

royaljack:
Thank you @evilhero !

PR #850, opened 2021-08-06
FIX: (#846) memory leak, FIX: removal of invalid ComicID's on startup, FIX: Updater invalid field patch

evilhero:
- FIX: (#846) memory leak on Manage Issues page
- FIX: on startup, make sure to remove problem IDs from comics table
- FIX: Updater would fail if ``LastUpdated`` field was not present


evilhero:
PR #850 will fix the loading of the manage issues page (it'll still be slow, but it will load and not jump the memory at least). 

Future update will revamp the page so that it loads dynamically instead of needing the entire dataset before being able to render.

barbequesauce (approved):

PR #905, opened 2021-10-07
IMP: Start of the major updates

evilhero:
This is the first of several major updates that are coming down the pipe over a short period of time. They will be big. There will be bugs (hopefully not lots), but I'm hoping that people testing this out in py3-dev, prior to hitting the master branch is enough of a workthru to get most of the bugs worked out.

This PR's big improvements are adding SSE event handlers, entire refactoring of the series detail page, new Queues for Refreshes and Adds, checking a specific folder prior to initiating a search, and literally a tonne of other stuff....

**Improvemnts**
- IMP: SSE event handlers added to GUI to allow for dynamic notifications throughout program regardless of screen - data will now be reflected in real-time (or very close to it). Only the index, history and series detail pages done currently...
-IMP: Refresh Queue and Add Series Queue implemented in order to queue up multipes in the background
- IMP: Folder Cache will be checked during each post-processing run in order to see if any matches are present before actual searching. (Folder Cache will be checked for subsequent new items at > 10 minutes since last post-processing run and held in memory)
- FIX: ``page_name`` field added to all html to denote whether or not dynamic updating is allowable (not all pages currently supported)
- IMP: changeBookType API endpoint added to allow for changing booktype of specific id
- IMP: directory delete API endpoint added option to delete physical directory (``directory_del`` = ``True``/``False``)
- IMP: refresh API endpoint now accepts comma delimited ``id`` list which flows thru the new Refresh Queue
- IMP: load_image function added to stream the image to UI instead of loading it (convert to webp for possible faster loading)
- IMP: Added publisher image for AWA
- IMP: comic detail page updated to dynamic loading table which allows for near instant page load regardless of size:
  - annuals section updated to show grouping by series, proper column headers to indicate groupings along with delete options
  - issue detail box redone to be loaded javascript side instead of python side
  - added ``metatdata_source`` and ``metadata_type`` to issue details box to display depending on issue metadata loaded
  - ``secondary`` graphic added to issue status to denote when an issue is located in a secondary destination folder
  - on initial page load, if have counts don't match to what's in folder count - will recheck files automatically on initial load prior to display
  - ``show cover date`` and ``show store date`` option available depending on what is being shown for issues of a series (will display in corresponding column)
  - filters are dynamic based on what status' are present in issue list (if status doesn't exist, will not show)
  - show filesize for image in MB, not in bytes
  - added series volume watermark to general information tab of series for better visibility
  - redid all buttons to run via javascript for better response control
  - fixed checkbox issue selection so annuals are selected separetely from issues (previous bug)
  - Archived color status changed
- FIX: storyarc and storyarc detail table layout was overlapping in some portions (adjusted css)
- FIX: removed ``explisit`` from ``findComic`` API endpoint
- FIX: if file returned ``Failed the read`` during metatagging, would incorrectly fail the item when it should be corrupt
- FIX: add additional test for more Print editions being incorrectly identified if One-Shot was tagged in the CV description field.
- FIX: correct calculation of image already existing when using alternate covers options and deciding on if a cover is new or not
- FIX: changed ``mode`` calls from unqueueissue function to ``smode``
- FIX: remove futurepull functions (not used)
- FIX: fixed auto-want having incorrect sql query
- FIX: corrected changing reading order sequence so it will resequence without error
- FIX: manual_metatagging would not properly resolve issues in secondary folder location
- FIX: manual post-processing location will now be saved (just the last run) upon a successful post-processing run

New variables added to config:
``MANUAL_PP_FOLDER`` #directory to default the manual post-processing option to
``FOLDER_CACHE_LOCATION`` # directory to scan during interval'd post-processing prior to searching for issue


barbequesauce (approved):

evilhero:
I'll re-do the PR with the extra commit that's not in it (was my fault for pushing lol)

PR #926, opened 2021-10-16
Fix magnet support for uTorrent

eb3095:
This fixes uTorrent and magnet links. addFile results in an error with magnet links, addurl is what is needed. I also added some more debug lines related to magnet as I was struggling with this before. IDE cleaned up trailing whitespace automatically.

I am currently using this, so it does work.

> Hope this is what you were looking for this time. Clean patch on a fresh fork of the python3-dev branch. Let me know if I am still misunderstanding this here.


evilhero:
Looks good - can't test as I don't run uTorrent, but don't see anything breaking in it.

Thanks for the PR!

PR #928, opened 2021-10-16
FIX: filter SABnzbd queue by nzo_id instead of search name

evilhero:
When searching the active SABnzbd queue would search nzo_ids by name using the search field (which searches nzb names), instead of using the nzo_ids field. 

Not sure if this was a new addition in SAB, or if it was always there - but this fixes endless looping of downloads when the queue is large.


barbequesauce (approved):
I wish my queue was large

PR #930, opened 2021-10-17
prevent AttributeError exception

djeebus:
after tagging some comics with ComicTagger, I found this in the log while trying to import many of the files:

```
Uncaught exception: Traceback (most recent call last):
File "/app/mylar3/mylar/logger.py", line 337, in new_run
old_run(*args, **kwargs)
File "/usr/lib/python3.8/threading.py", line 870, in run
File "/app/mylar3/mylar/webserve.py", line 5022, in preSearchit
if 'annual' in getiss.lower():
AttributeError: 'NoneType' object has no attribute 'lower'
```

Seems like `result['IssueNumber']` can occassionally be `None`, this seemed like the easiest way to prevent the error.

### DIFF
--- a/mylar/webserve.py
+++ b/mylar/webserve.py
@@ -5017,7 +5017,7 @@ class WebInterface(object):
                     else:
                         #logger.fdebug('result: %s' % result)
                         comicstoIMP.append(result['ComicLocation']) #.decode(mylar.SYS_ENCODING, 'replace'))
-                        getiss = result['IssueNumber']
+                        getiss = result['IssueNumber'] or ''
                         #logger.fdebug('getiss: %s' % getiss)
                         if 'annual' in getiss.lower():
                             tmpiss = re.sub('[^0-9]','', getiss).strip()


evilhero:
First, thanks for the PR.

However, this won't change the fact that importing files without an issue number is still broken and will be addressed when the refactor goes live. 

This is just a temporary measure to fix the current state of the importer (meaning this will introduce additional bugs in regards to allowing this since the rest of the code hasn't been updated).  Future PR's in regards to the import process may not be pushed until such time as the refactor goes live as then we're dealing with multiple variations of the same code and maintaining different variations becomes a tad confusing.

Just warning that this may not address the issues if you're importing files with no issue numbers.

Also this is against the master branch. 

We only push PR's to the python3-dev branch. You'll have to create a new, clean PR with just the changes in this PR (meaning don't just pull the differences btwn master and python3-dev and update this PR - you'll see an additional commit labelled as python3-dev merge or something if you do).

djeebus:
gotcha, thanks for the quick response! will keep in mind the correct branch to point to in the meantime.

So I should probably just sit tight until the refactor gets into master? I can live with that, thanks again!

evilhero:
Well I mean if the PR works for you - that's one thing. We've known about the importer being broken for a while for filenames that don't contain an issue number, thus the refactor process (plus the importer isn't the greatest since the code has been around for like 8+ years).

We're still talking a few months (maybe less?) out for the importer most likely - mainly due more to the number of bugs that have been popping up and having to deal with those. The importer code-base is mostly done, but it needs heavy testing prior to being released along with a few more code updates - so it's not a hopeless process.

If you want to do the clean PR, we'll merge it and see where it goes from there. I mean if it saves me doing additional work in the refactor then great, but it's about being consistent with the refactor and the updated code changes we're pushing now as well.

PR #946, opened 2021-11-06
Filechecker none bug

mlsmithjr:
I get an exception on line filechecker.py:143 when self.watchcomic is None.  The all() function does not short circuit like it should so I reverted it to a simple 'and' comparison.


evilhero:
Please provide a debug log showing what's happening. Those checks are in place for alot of reasons so deconstructing it, while it might work for your use case, may break additional things in other aspects.

Also, please make sure you update your branch prior to creating the PR. We can't merge it up due to the way we merge if there are previous merge details included, so we ask for a clean PR which has just the changes and that's all.

PR #982, opened 2021-12-24
certgen.py fix per pep 3113

btxgit:
- Changed the two python2-style unpacked tuple parameters into python3 compatible parameters.
- Oddly, there was a separate commit to my repo that's still valid, which skips over symlinks that aren't pointing at valid files (otherwise the getsize() call woudl throw an exception.


evilhero:
Can you do the git pull prior to doing the PR? That python3-dev merge in your PR will mess up when we go to merge into master.

We prefer to keep the PR's isolated to just the affected changes, not previous commits.

Like you didn't know alrwsfh, thanks for helping (again)!

PR #983, opened 2021-12-25
Update certgen.py per PEP 3113 (tuple parameters)

btxgit:
Fixed up createCertificate parameters as per PEP 3113


barbequesauce (approved):

barbequesauce:
Thanks for the contribution (as always)!

PR #999, opened 2022-01-12
Here. Hold my sink. And all of this.

evilhero:
IMP: vast improvement in RSS searching time (refactored)
IMP: maintenance-mode GUI mode with progress indicators, resume and isolated web GUI during upgrade path when upgrading db (as well as progress in CLI)
IMP: DDL downloads will get the ``[__ISSUEID__]`` attached temporarily to filename to lessen chance of problems with naming problems
IMP: Break out search matching functionality from search module into it's own module
IMP: (refactor) Breaking out allows DDL items to be verified for matches during queries saving hits & time
IMP: Priority Alternate Search Name added - use !! instead of ## to denote priority vs normal
IMP: Provider-search delay (global setting) is now honoured on a per provider basis and is persistent across restarts.
IMP: Skipped2Wanted will now take both issues & annuals into account for a series
IMP: Break out DDL into DDL(GetComics) to allow for future additions (config auto-update to v12)
IMP: Force Enable DDL to mirror DDL(GetComics) when enabling/disabling via GUI since only option available atm
IMP: Flaresolverr support added (specifically for GC usage only)
IMP: Cookie receipts when using GC
IMP: Extra loop for tpb/gn/one-shot/annual inclusion when using DDL to account for issues with no issue number
IMP: ``ddl_query_delay`` available to delay search queries to GC (default 15s) and only affects search queries
IMP:  DDL downloads will now temporarily append the issueid to the filename after downloading to improve post-processing
FIX: Annual integration broken and not auto-adding annuals outside of manually integrating them
FIX: Error when post-processing issues with the ``[__ISSUEID__]`` included in the filename for direct issue processing.
FIX: ``Add to Readinglist`` and ``Download`` icons & functionality restored to series detail page
FIX: Deleting series in GUI via seriesdetail page would sometimes throw a 500 error
FIX: Make sure to return a null dataset instead of just null when no searchresults returned
FIX: Remove some extra logging lines needed during testing
FIX: Search-4-missing would not throw a popup after queue submission
FIX: Forgot to include One-Shot as part of the TPB/GN/HC json return to show possibility of having no issue numbers
FIX: Status colours on Scheduler Management page
FIX: Extra spacing at bottom of Edit Settings tab on series detail page causing tab size to be bigger than necessary
FIX: Try to adjust Folder Name creation to retain $VolumeN when series is a non-print edition

### DIFF
--- a/Mylar.py
+++ b/Mylar.py
@@ -25,7 +25,7 @@ sys.path.insert(1, os.path.join(os.path.dirname(__file__), 'lib'))
 
 import mylar
 
-from mylar import webstart, logger, filechecker, versioncheck, maintenance
+from mylar import webstart, logger, filechecker, versioncheck, maintenance, maintenance_webstart
 
 import argparse
 
@@ -103,7 +103,6 @@ def main():
     # Set up and gather command line arguments
     parser = argparse.ArgumentParser(description='Automated Comic Book Downloader')
     subparsers = parser.add_subparsers(title='Subcommands', dest='maintenance')
-    parser_maintenance = subparsers.add_parser('maintenance', help='Enter maintenance mode (no GUI). Additional commands are available (maintenance --help)')
 
     #main parser
     parser.add_argument('-v', '--verbose', action='store_true', help='Increase console logging verbosity')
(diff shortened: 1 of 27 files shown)


barbequesauce (approved):
Buckle up buttercup

Issue #1006, opened 2022-01-23
More Descriptive/Accurate Notifications


Nymarra:
**Describe the request**
A clear and concise description of what the request is.

Currently notifications are barebones and lack some key information, failing to convey partial failures of post-processing (file moved, but failed to convert, for example). More verbose discord messages that trigger on various conditions/states could be helpful for identifying these. This would be especially useful when identifying bad downloads, when a file failed to convert, it's likely because it was a bad file to begin with.

As a minor nit, it would also be more correct to state post-processing is completed without mentioning downloads due to manual post-processing, but this is a wording nit.

Additionally, the current notifications are sent before conversion is attempted, meaning it couldn't catch failed conversions anyway. As a stopgap if more verbose messages aren't something you would like to pursue, moving it later in the post-processing flow could flag that an error occurred without being overly specific.

**How did you come up with the idea?**
Have you seen this request implemented somewhere else? An example or two might speed along development

I was observing the logs while simultaneously watching discord for notifications, and noticed a notification get sent (both in logs and discord, then conversion attempted and failed. The message I got via discord was about an issue being downloaded and post-processed correctly. 

It made me wonder how many comics were marked as "post processed" that are actually bad files in my library. Having an alert about a potential error would be massively helpful.

**Environment (please complete the following information):**
 - OS: [e.g. Windows10] MacOS
 - Branch / commit: [e.g. Dev/38aksi1] Dev/07180d18e98e9c6160a1603798168b595b95b25e

**Additional information**
Add any other information about the request here.
Link to a log output showing this error in action can be found in the discord at https://discord.com/channels/826493118106697758/826498038885122149/933926471188230165

barbequesauce:
@misleadingrhino given that was your PR, can you take a crack at addressing it please?

Nymarra:
Resolved in PR 1009

PR #1009, opened 2022-01-23
Add notifications for errors with the metatagger

misleadingrhino:
Added a notify function to the metatagger. Also modified discord notifications to add yellow error embed for these errors. Other notifications **should** work, but I wasn't able to test them. 
image


misleadingrhino:
#1006

barbequesauce (approved):

PR #1115, opened 2022-03-19
FIX: CT would fail to metatag issues due to CT-UA problems

evilhero:
- IMP: change out urllib for requests in CT, use params instead of string url
- FIX: CT would fail to metatag issues due to invalid UA (auto-generate random UA)
- IMP: log CT errors as unsuccessful when metatagging and do not complete as if was successful
- IMP: bump CT version to 1.3.5


lib/comictaggerlib/comicvinetalker.py
```diff
@@ -134,13 +132,12 @@ def parseDateStr(self, date_str):
...
 
         try:
-            test_url = self.api_base_url + "/issue/1/?api_key=" + \
-                key + "&format=json&field_list=name"
-            resp = urllib.request.urlopen(test_url, context=self.ssl)
-            content = resp.read()
-    
-            cv_response = json.loads(content.decode('utf-8'))
-    
+            test_url = self.api_base_url + "/issue"
```
btxgit, line 135:
```suggestion
            test_url = self.api_base_url + "/issue/1/"
```
Missing the  /1/.

lib/comictaggerlib/comicvinetalker.py
```diff
@@ -233,10 +230,15 @@ def searchForSeries(self, series_name, callback=None, refresh_cache=False):
...
 
-        search_url = self.api_base_url + "/search/?api_key=" + self.api_key + "&format=json&resources=volume&query=" + \
-            query_string + \
-            "&field_list=name,id,start_year,publisher,image,description,count_of_issues&limit=100"
-        cv_response = self.getCVContent(search_url + "&page=1")
+        search_url = self.api_base_url + "/search"
+        params = {'api_key': self.api_key,
+                  'format': 'json',
+                  'resources': 'volume',
+                  'query': query_string,
```
btxgit, line 237: I am concerned this will double encode because this query_string has already been escaped wtih quote_plus.  I would comment out the query_string = urllib.parse.quotplus(query_string.encode("utf-8")) line and just leave the escaping to requestss params.

lib/comictaggerlib/comicvinetalker.py
```diff
@@ -411,13 +417,15 @@ def fetchIssuesByVolumeIssueNumAndYear(
 
         issue_number = urllib.parse.quote_plus(str(issue_number).encode("utf-8"))
 
-        filter = "&filter=" + volume_filter + \
-            year_filter + ",issue_number:" + issue_number
+        filter_line = '%s%s,issue_number:' % (volume_filter, year_filter, issue_number)
```
btxgit, line 420:
```suggestion
        filter_line = '%s%s,issue_number:%s' % (volume_filter, year_filter, issue_number)
```
There's 3 args, but there were only 2 format strings.

btxgit (changes requested):
At least one of these will cause a fatal exception, so this should not be approved yet.

mylar3 (approved):

PR #1118, opened 2022-03-20
Manual DDL GetComics.info link

slarrain:
This a small new feature for Mylar3 that I would like to add.

Many times, I do not get the right Issue snatched or downloaded through the manual or wanted search. And some of those times, I know exactly which is the right URL for fetching that comic on GetComics.info.
So I wanted to be able to add a button on the issues table, that would show up if the issue hasn't been snatched or downloaded, and that the presssing of said button would open a dialog with a text input that would allow us to write down the URL for that issue.
image
Then, the `SEARCH-QUEUE`would start a custom search, that would skip all the actual finding of the URL (since we already know the correct one) and would straight away fetched the comic file.
This PR aims to implement that functionality.

I tried to edit or change the Mylar3 base code as little as possible. It was actually quite  a challenge getting to understand how does the `SEARCH-QUEUE` works, which parts can be skipped, and which cannot.

Review and comments would be very welcomed. This was my first time using JS, so please don't hold it against me.

DEMO VIDEO: 
https://www.youtube.com/watch?v=zGl7SMOzI4o


barbequesauce:
We all hate rebase....

@evilhero pending your feedback please

slarrain:
Just a friendly reminder to @evilhero about this PR

ducky0518:
I agree this would be a welcomed addition for those of us running Mylar on a remote system. Thanks!

cardonator:
This is an absolutely required feature, IMO. Managing comics that Mylar fails to find or download properly is an exercise in pain.

gotson:
Would love to see this added. At the moment the way around it is to download, place the file in a folder on the mylar server and run manual processing, but that doesn't always work.

theotocopulitos:
I'd love to see this integrated in mylar3

agiacalone:
I have a sizable comic collection and this feature is desperately needed.

barbequesauce:
@slarrain have your ears been burning? We've been talking about you...
Apologies we sat on this so long. We see value in this change but there's been a fair amount of change since this was submitted; any chance you could submit a new PR based to latest python3-dev?  I know at least one other dev has some suggested changes as well.

falo2k:
Just a few suggestions for a resubmit:

- I wouldn't put your personal choice of venv path into .gitignore.
- Standardise that action string to lowercase for consistency
- I'm not a fan of putting additional GC specific code into helpers.py.  That should be within the GC module.
- The handling for incorrect urls seems off.  How exactly is that expected to behave to an end user?  At the moment you catch the exception, throw an ERROR log, and then return None as an ID and carry on with the URL regardless.

barrelltitor:
I would keep this logic separate from the actual search queue. Better extract the logic you need into a function and use that both in the search queue and in your code

Making the search queue handling more complex is not ideal as right now there really needs to be some modularization done

PR #1119, opened 2022-03-20
Compare length of result substring against length of comic name

kcpants:
Prior to commit cdddddbc93af171f9c90bc84f0a66613f685f71f this condition was `if len(subs) >= len(seriesname)`. I believe it is meant to be a sanity test to determine if the substring likely contains the series name and should test whether the substring is longer than the series name. As written I believe it is testing whether the substring in the search result is longer than the _number of fields_ in the comicName.

Without this fix the first field of the search result _2021-09-01 Indie [25/31] - "The Many Deaths of Laila Starr 05 (of 05) (2021) (digital) (Son of Ultron-Empire).cbr" yEnc_ was identified as the title. E.g.
```
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.72 : SEARCH-QUEUE : checking search result: 2021-09-01 Indie [25/31] - "The Many Deaths of Laila Starr 05 (of 05) (2021) (digital) (Son of Ultron-Empire).cbr" yEnc
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.90 : SEARCH-QUEUE : sub: 2021-09-01 Indie [25/31] - 
```
With this fix the second field is correctly identified as the title name. E.g.
```
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.90 : SEARCH-QUEUE : sub: The Many Deaths of Laila Starr 05 (of 05) (2021) (digital) (Son of Ultron-Empire).cbr  ComicName: The Many Deaths of Laila Starr
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.142 : SEARCH-QUEUE : comsize_b: 4322567320-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.155 : SEARCH-QUEUE : size given as: 41.2 MB
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.235 : SEARCH-QUEUE : store date used is : 2021-04-2120-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.236 : SEARCH-QUEUE : date used is : 2021-04-21
20-Mar-2022 20:57:30 - DEBUG   :: mylar.checker.288 : SEARCH-QUEUE : usedate: 2021-04-21
```

### DIFF
--- a/mylar/search_filer.py
+++ b/mylar/search_filer.py
@@ -90,7 +90,7 @@ class search_check(object):
                     logger.fdebug('sub: %s' % subs)
                     try:
                         if (
-                            len(subs) >= len(ComicName.split())
+                            len(subs) >= len(ComicName)
                             and not any(d in subs.lower() for d in except_list)
                             and bool(_digits.search(subs)) is True
                         ):


evilhero:
You'll have to do this against the python3-dev branch, as we never commit directly to the master branch. Once against the correct branch, we can merge it. 

But yes, you are correct in that it's taking the wrong value with the ``split()`` being included. 

The idea behind it, is that section of code would be a really basic way of getting rid of the extra "stuff" around the filename that happens when nzb providers are lazy and don't set up proper filters/regex's for comics on their system, which results in headers being included in the search results.

This snippet was taken from the findcomicfeed module which does the exact same thing for experimental search since that deals in raw headers (and no split in there). 

Also, thanks for the PR - don't think this would have been caught otherwise as it would only affect users who have a provider returning the raw headers (which obviously is a very small subset).

Cheers!

PR #1151, opened 2022-04-10
Initial support for Mattermost notifications

Dracon23:
Simple notifications based off the Slack notifier.

Preview

A followup with extended notifications (using attachments) will follow once I feel a little more at home with Mattermost's payload format and Mylar's notification API.


mylar/webserve.py
```diff
@@ -7815,6 +7818,17 @@ def testslack(self, webhook_url):
...
+    
+    def testmattermost(self, webhook_url):
+        mattermost = notifiers.MATTERMOST(test_webhook_url=webhook_url)
+        result = mattermost.test_notify()
+
+        if result == True:
+            return "Successfully sent Mattermost test -  check to make sure it worked"
+        else:
+            logger.warn('Test variables used [WEBHOOK_URL: %s][USERNAME: %s]' % (webhook_url, username))
+            return "Error sending test message to Slack"
```
BRTPOB, line 7830: I don't think you meant to refer to Slack here. Might want to fix this.

mylar/PostProcessor.py
```diff
@@ -3425,6 +3425,10 @@ def sendnotify(self, series, issueyear, issuenumOG, annchk, module, imageFile):
                 slack = notifiers.SLACK()
                 slack.notify("Download and Postprocessing completed", prline2, module=module)
 
+            if mylar.CONFIG.MATTERMOST_ENABLED:
+                mattermost = notifiers.MATTERMOST()
+                mattermost.notify("New comic available", prline2, module=module)
```
BRTPOB, line 3430: We probably want to keep the notification standardized to what the other notifications read as.

BRTPOB: Also, is there a reason we're not including the image here? Does Mattermost not support images in messages?

Dracon23: > Also, is there a reason we're not including the image here? Does Mattermost not support images in messages?

Not in the message body, as far as I could glean from their webhook documentation. I am currently working out how to properly include it into the attachment.

mylar/webserve.py
Dracon23: Whoop, missed that one. Fixed in 28ce225549331e5342ca4ca542c010dbe214a7f1

mylar/PostProcessor.py
Dracon23: > We probably want to keep the notification standardized to what the other notifications read as.

Makes sense. Adjusted in 589619eec388b19f2bb501f7ee243b33aaae8a23

barbequesauce (approved):

PR #1192, opened 2022-04-30
Reading lists

jakem742:
Creates a "Generate Reading List" button which creates a CBL file containing an ordered list of the issues in the Story Arc.

Notes:

- Creates a "ReadingLists" folder at the root of the data folder (next to "StoryArcs", not within). There isn't any guarantee the user has enabled "Arcs in StoryArc Directory", so additional checks would need to be made before this could be placed within the StoryArcs directory.
- The SQL query excludes all issues that have been manually deleted by the user in Mylar

### DIFF
--- a/mylar/webserve.py
+++ b/mylar/webserve.py
@@ -4924,6 +4924,83 @@ class WebInterface(object):
         raise cherrypy.HTTPRedirect("detailStoryArc?StoryArcID=%s&StoryArcName=%s" % (storyarcid, storyarc))
     importReadlist.exposed = True
 
+    def GenerateCBLReadingList(self, StoryArcID=None):
+
+        myDB = db.DBConnection()
+
+        if StoryArcID:
+            ReadingArc = myDB.select("SELECT * FROM storyarcs WHERE StoryArcID=? AND (Manual!='deleted' OR Manual IS NULL) ORDER BY ReadingOrder", [StoryArcID])
+
+        if ReadingArc is None:
+            logger.info("Couldn't find Story Arc to use for reading list!")
+        else:
+            arcName = ReadingArc[0]['StoryArc']
+            logger.fdebug('Generating CBL file for ' + arcName)
+
+            # !Replace root_path variable
+            readingListDest = os.path.join(mylar.CONFIG.DESTINATION_DIR, 'ReadingLists')
+            filename = arcName + ".cbl"
+            fullPath = os.path.join(readingListDest, filename)
+
+            if not os.path.exists(readingListDest):
+                logger.fdebug('making directory: %s' % readingListDest)
+                os.mkdir(readingListDest)
+
+            logger.fdebug('Opening file: ' + fullPath)
+
+            with open(fullPath, 'w') as file:
+
+                line1 = "<?xml version=\"1.0\"?>"
+                line2 = "<ReadingList xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">"
+                line3 = "<Name>" + escape(arcName) + "</Name>"
+                line4 = "<Books>"
+
+                file.writelines([line1, line2, line3, line4])
+
+                #For each issue in arc
+                for issue in ReadingArc:
+
+                    #Set default values to blank string
+                    fileName = seriesName = issueNumber = volNumber = issueYear = ""
+
+                    if issue['Location'] is not None:
+                        filePath = issue['Location'].split("/")
+                        #lastItemIndex = len(filePath) - 1
+                        fileName = escape(filePath[-1])
+                        
+                        #Strip reading order numbers from start of filename if present
+                        regex = re.compile("^\d{3}-(.*?)\.(cbz|cbr)")
+                        match = regex.match(fileName)
+                        if match != None:
+                            strippedName = '{}.{}'.format(*match.groups())
+                            fileName=strippedName
+
+                    if issue['ComicName'] is not None:
+                        seriesName = escape(str(issue['ComicName']))
+                    if issue['IssueNumber'] is not None:
+                        issueNumber = escape(str(issue['IssueNumber']))
+                    if issue['Volume'] is not None:
+                        volNumber = escape(str(issue['Volume']))
+                    if issue['SeriesYear'] is not None:
+                        issueYear = escape(str(issue['SeriesYear']))
+                    if issue['IssueID'] is not None:
+                        issueCVID = escape(str(issue['IssueID']))
+
+                    file.write("<Book Series=\"" + seriesName + "\" Number=\"" + issueNumber + "\" Volume=\"" + volNumber + "\" Year=\"" + issueYear + "\">")
+                    file.write("<FileName>" + fileName + "</FileName>")
+                    file.write("<CVID>" + issueCVID + "</CVID>")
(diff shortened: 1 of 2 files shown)


mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+        myDB = db.DBConnection()
+
+        if StoryArcID:
+            ReadingArc = myDB.select("SELECT * FROM storyarcs WHERE StoryArcID=? AND (Manual IS NOT 'deleted' OR Manual IS NULL) ORDER BY ReadingOrder", [StoryArcID])
+
+            if ReadingArc is None:
+                logger.info("Couldn't find Story Arc to use for reading list!")
+            else:
+                arcName = ReadingArc[0]['StoryArc']
+                logger.fdebug('Generating CBL file for ' + arcName)
```
evilhero, line 4938: shouldn't use ``+`` to format the value. Use ``%s``  method. Using the ``+`` method will generate traceback errors if it's a None value.

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+
+                if not os.path.exists(readingListDest):
+                    logger.fdebug('making directory: %s' % readingListDest)
+                    filechecker.validateAndCreateDirectory(readingListDest, True)
+
+                logger.fdebug('Opening file: ' + fullPath)
+
+                with open(fullPath, 'w') as file:
+
+                    line1 = "<?xml version=\"1.0\"?>"
```
evilhero, line 4953: Need to add the ``encoding=\"UTF-8\"`` so that it's encoded properly. Escaping items in a text file is unnecessary nowadays as indicating what the xml is encoded at is enough for the clients.

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                    line2 = "<ReadingList xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">"
+                    line3 = "<Name>" + escape(arcName) + "</Name>"
```
evilhero, line 4955: escape in this reference will cause a traceback when generating the cbl. ``escape`` isn't loaded anywhere and so it cannot be referenced. There is no need to escape, with the encoding set to utf-8. Just remove it and do not use the + to join the   values - do it like ``"<Name>%s</Name>" % (arcName)``

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                    file.writelines([line1, line2, line3, line4])
+
+                    #For each issue in arc
+                    for issue in ReadingArc:
+
+                        #Set default values to blank string
+                        fileName = seriesName = issueNumber = volNumber = issueYear = ""
+
+                        if issue['Location'] is not None:
+                            filePath = issue['Location'].split(os.path.sep)
```
evilhero, line 4967: It's easier, safer and OS agnostic if you just use os.path.basename instead of doing the path.sep -1 thing. ( ``filepath = os.path.basename(issue['Location'])`` )

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                            #lastItemIndex = len(filePath) - 1
+                            fileName = escape(filePath[-1])
```
evilhero, line 4969: Not needed with the above os.path.basename - need to remove.

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                            
+                            #Strip reading order numbers from start of filename if present
+                            regex = re.compile("^\d{3}-(.*?)\.(cbz|cbr)")
+                            match = regex.match(fileName)
+                            if match != None:
+                                strippedName = '{}.{}'.format(*match.groups())
+                                fileName=strippedName
+
+                        if issue['ComicName'] is not None:
+                            seriesName = escape(str(issue['ComicName']))
```
evilhero, line 4979: escape isn't needed - just assign it as the value without the string (you could keep it, but it's not needed).

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        if issue['IssueNumber'] is not None:
+                            issueNumber = escape(str(issue['IssueNumber']))
```
evilhero, line 4981: escape isn't needed - just assign it as the value without the string (you could keep it, but it's not needed).

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        if issue['Volume'] is not None:
+                            volNumber = escape(str(issue['Volume']))
```
evilhero, line 4983: escape isn't needed - just assign it as the value without the string (you could keep it, but it's not needed).

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        if issue['SeriesYear'] is not None:
+                            issueYear = escape(str(issue['SeriesYear']))
```
evilhero, line 4985: escape isn't needed - just assign it as the value without the string (you could keep it, but it's not needed).

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        if issue['IssueID'] is not None:
+                            issueCVID = escape(str(issue['IssueID']))
```
evilhero, line 4987: escape isn't needed - just assign it as the value without the string (you could keep it, but it's not needed).

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+
+                        file.write("<Book Series=\"" + seriesName + "\" Number=\"" + issueNumber + "\" Volume=\"" + volNumber + "\" Year=\"" + issueYear + "\">")
```
evilhero, line 4989: this will cause problems with using ``+`` to join the values. Should be :
``file.write("<Book Series=\"%s\" Number=\"%s\" Volume=\"%s\" Year=\"%s\">" % (seriesName, issueNumber, volNumber, issueYear))``

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        file.write("<FileName>" + fileName + "</FileName>")
```
evilhero, line 4990: same as above, join with ``+`` is bad. ``file.write("<FileName>%s</FileName>" % fileName)``

mylar/webserve.py
```diff
@@ -4924,6 +4924,83 @@ def importReadlist(self, filename):
...
+                        file.write("<CVID>" + issueCVID + "</CVID>")
```
evilhero, line 4991: as above. Should be ``file.write("<CVID>%s</CVID>" % (issueCVID))``

evilhero:
Also, a clean, up-to-date PR would be required once those changes are implemented (you're missing a bunch of previous commits) that would get hammered if we were to merge up. 

By clean, I mean that we will only merge PR's that consist of just the changes you're making - no previous commits, merges by anyone else is to be within the PR itself.

barbequesauce:
Closing due to staleness...

PR #1241, opened 2022-06-17
Detect running under Kubernetes as a docker install

ganeshjp:
Currently, running the linuxserver/mylar3 docker container as kubernetes pod is not detected as a container/docker install causing a traceback. This PR adds a check for the environment variable KUBERNETES_SERVICE_HOST to the docker install mode detection logic.

### DIFF
--- a/mylar/versioncheck.py
+++ b/mylar/versioncheck.py
@@ -183,7 +183,7 @@ def getVersion(ptv):
     else:
 
         d_path = '/proc/self/cgroup'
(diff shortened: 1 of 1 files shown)


evilhero:
First, thanks for the PR!

However, You'll have to resubmit this against the python3-dev branch as we only push items to the dev branch. Just make sure you're up-to-date prior to doing the PR.

(Or we can just make the change for you and do up a PR if you don't have the time / inclination any longer)

Cheers
(Closing this as it's against the wrong branch)

PR #1262, opened 2022-07-27
FIX: DDL Queue buttons not displayed if the Active download is in a Failed state

evilhero:


barbequesauce (approved):

barbequesauce:
boring branch name.

PR #1313, opened 2022-12-06
Tweaked Issue ID parsing on import to handle additional numbers in metadata Notes field

falo2k:
My attempt to squash a very niche bug.  I had a bunch of comics with some very old scraper comments where there were numbers (in this case dates) later in the Notes field than the CVDBID / Issue ID.  The current parsing method did not account for this, so would not recognise the Issue ID correctly.  This change just uses a different regex approach to look for the first set of consecutive numbers following either "Issue ID" or "CVDB".  I've allowed it to ignore any characters between those prefixes and the ID in case there's some odd spacing/separators for some reason.

Example of Notes field that would fail on parsing the ID:
`Scraped metadata from ComicVine [CVDB196459] on 2011.05.20 08:41:41.`

The old method would strip this down to `196459 2011 05 20 08 41 41` and decide it wasn't a digit field.

### DIFF
--- a/mylar/librarysync.py
+++ b/mylar/librarysync.py
@@ -363,18 +363,12 @@ def libraryScan(dir=None, append=False, ComicID=None, ComicName=None, cron=None,
                                 # if used by ComicTagger, Notes field will have the IssueID.
                                 issuenotes = issueinfo['metadata']['notes']
                                 logger.fdebug('[IMPORT-CBZ] Notes: ' + issuenotes)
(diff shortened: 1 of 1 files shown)


evilhero (approved):

evilhero:
Good to go and merged into python3-dev.  Thanks for the PR ! 

Cheers

falo2k:
np.  Thank you for all your work!

Issue #1179, opened 2022-04-22
Cant save via Web GUI


mjd515:
i cant get the web GUI to save it always displays an an error occurred. i can edit the file manually with a text editor, and the changes will replicate after a reboot. 

i am using docker/docker compose on linux to load the application, i am using the linuxserver nightly image 

i have tried chmoding the file to 777 as well as verifying my GUID and PUID is set correctly 

i have attached a link to the carepackage 

https://mega.nz/file/5pA0CKTC#4VcnU3N6LyTYBjqWvi0S0wYdy_UzPd7CmUQh0RutTpM

DieselTech:
I'm not seeing anything stand out in the carepackage. You mention that it shows an error occurred when trying to save, what specifically is that error and do you happen to have screenshots of it?

mjd515:
I can screen shot it but it literally just says an error occured in a red box at the bottom. And then there isn't anything in the logs even after turning on debug

barbequesauce:
How about in the JavaScript console /developer tools?

mjd515:
[mylar.mdaps.net-1650728723416.log](https://github.com/mylar3/mylar3/files/8547532/mylar.mdaps.net-1650728723416.log)
thats the error i get in the console i also have a text file of the out put of the erron from the URL as a stack trace as well as node js
[new 1.txt](https://github.com/mylar3/mylar3/files/8547537/new.1.txt)

evilhero:
Check your nginx logs for an error. 

If I recall from another post along time ago (that I cannot find for the life of me), because Mylar sends the complete config in one url fetch, it can cause nginx to barf if the length is too long due to ``large_client_header_buffers`` being set to a lower than default option. 

You may have to specify that value for your mylar endpoint in your nginx config.

mjd515:
i use traefik and not ngix unless the mylar app uses ngix in the background?

barbequesauce:
No it doesn't. I believe there are similar config options in traefik...

mjd515:
ok ill check it out. i didnt see any errors in traefik logs but im sure thats probably the case. thanks! ill close the issue if i figure it out or come back with findings if no change.

evilhero:
Yeah my point being it's not really a Mylar issue as far as being able to save. There's some config option that needs to be enabled or disabled so that the large url can be saved proeprly from whatever proxy manager you're using. 

If you were to try and save the config outside of the reverse proxy, it would be able to save the config.

[...]

PR #1322, opened 2023-02-05
FIX (#1179) Change ajax method from get to post

barbequesauce:
FIX (#1179) Change ajax method from get to post

### DIFF
--- a/data/js/script.js
+++ b/data/js/script.js
@@ -194,6 +194,7 @@ function doAjaxCall(url,elem,reload,form) {
 	// Ajax Call
 	$.ajax({
 	  url: url,
+	  method: POST,
 	  data: dataString,
 	  beforeSend: function(jqXHR, settings) {
 	  	// Start loader etc.


barbequesauce:
@evilhero small note on this - we have a few versions of jQuery running around our deployment, two of three meet the minimum requirements for this (jQ 1.9) but one doesnt....

Screenshot 2023-02-05 at 4 30 12 PM

evilhero:
@barbequesauce I believe I came up with a method to call the POST specifically instead of changing the ajaxcall on a global scale, which might cause some problems if it was expecting GET. 

Basically, just added another variable to the doAjaxCall function called ``method``. If it's not specified it will default to ``get``. In the config.html we just add ``post`` as the method for the doAjaxCall and that's it.

```js
+++ b/data/js/script.js
@@ -138,7 +138,7 @@ function showMsg(msg,loader,timeout,ms) {
        }
 }

-function doAjaxCall(url,elem,reload,form) {
+function doAjaxCall(url,elem,reload,form,method) {
        // Set Message
        feedback = $("#ajaxMsg");
        update = $("#updatebar");

@@ -154,6 +154,10 @@ function doAjaxCall(url,elem,reload,form) {
        var formID = "#"+url;
        if ( form == true ) {
                var dataString = $(formID).serialize();
+                if ( !method ) {
+                    method = 'get';
+                }
        }
```

```html
+++ b/data/interfaces/default/config.html
@@ -1712,7 +1712,7 @@
              </tr>
         </table>
     </div>
-        <input type="button" value="Save Changes" onclick="doAjaxCall('configUpdate',$(this),'tabs',true);return false;" data-success="Changes saved successfully">
+        <input type="button" value="Save Changes" onclick="doAjaxCall('configUpdate',$(this),'tabs',true,'post');return false;" data-success="Changes saved successfully">
         <div class="message">
                <p><span class="ui-icon ui-icon-info" style="float: left; margin-right: .3em;"></span>Web Interface changes require a restart to take effect</p>
        </div>
```

PR #1326, opened 2023-02-22
Updated to use glob + logging for output

DieselTech:
Using glob to walk the files now instead of find. Shaved 2+ minutes off running time compared to old script on a library of 100k files. 

Original:
real    3m27.720s
user    1m9.223s
sys     0m14.013s

Updated:
real    1m12.167s
user    1m7.907s
sys     0m4.241s

Also just adds basic texts files for the output so you don't have to watch the terminal output.

### DIFF
--- a/utilities/look4untaggedcbzs.py
+++ b/utilities/look4untaggedcbzs.py
@@ -1,14 +1,22 @@
+from lib2to3.pgen2.token import NEWLINE
 import sys
 import zipfile
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


evilhero:
Have to make it a clean PR sir. There's a python3-dev merge commit that needs to be removed. You can do this by first updating the python3-dev branch, then create the pull branch. Then when you do the PR it will be just the single commit.

DieselTech:
Nuked

PR #1327, opened 2023-02-23
Update look4untaggedcbzs.py

DieselTech:
Using glob to walk the files now instead of find. Shaved 2+ minutes off running time compared to old script on a library of 100k files.

Original:
real 3m27.720s
user 1m9.223s
sys 0m14.013s

Updated:
real 1m12.167s
user 1m7.907s
sys 0m4.241s

Also just adds basic texts files for the output so you don't have to watch the terminal output.


evilhero (approved):
good to go

PR #1331, opened 2023-03-11
Bugfixes for ComicID not being read from CV notes field

billroemhild:


evilhero:
While I understand what you're doing - is the overall intent just to handle traceback errors when the notes field returns an invalid item? 

The regex has been modified in the python3-dev branch but has not been merged into master, so you may have to accomodate your code to what is already in dev. 

This hasn't come up as of yet so I'm assuming it's in relation to what you have in your notes? Just needing some justification / reason as to why this is necessary outside of being able to handle a crash (which is fine, again looking for context). 

Also, we never merge into master, always python3-dev. So this will be closed due to setting the PR against the wrong branch. You'll have to redo a clean PR against the dev branch and then we can look into merging or addressing any concerns. 

Cheers.

billroemhild:
My mistake of not working in python3-dev branch.

My main goal here was to get better results of importing by using the ComicID.  After this fix I had 20% so better results while importing.

billroemhild:
I'd been working on a better workflow for new series where I use comictagger 1.5.5 to find the ComicID before handing off to Mylar.  Problem was that the ComicID tag then wasn't being used in Mylar.

evilhero:
Unless CT has changed recently, they don't write the comicid to the comicinfo.xml file, they only write the issueid into the notes field. Mylar is only built to use the issueid in the field - if the comicid is inserted it either won't use it, or will try to use it as the issueid and fail miserably. Mylar will use the issueid and reverse lookup the comicid in order to gather the relevant series info. 

Maybe you can verify what's put in the Notes field after tagging with CT, and let us know verbatim what it is. I have a feeling that they might have changed the formatting and it may be throwing off some aspects of the imporr (ie. It's specifically looking for ``Issue ID XXXXXX``)

PR #1334, opened 2023-03-15
Allow CBR files to be scanned for ComicInfo.xml during import

billroemhild:
Calling function log entries "IMPORT-CBZ" should probably be cleaned up

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -320,14 +320,22 @@ def rename_param(comicid, comicname, issue, ofilename, comicyear=None, issueid=N
                     if all([chkissue is None, annualize is None, not mylar.CONFIG.ANNUALS_ON]):
                         chkissue = myDB.selectone("SELECT * from annuals WHERE ComicID=? AND Issue_Number=? AND NOT Deleted", [comicid, issue]).fetchone()
 
-                if chkissue is None:
-                    #rechk chkissue against int value of issue #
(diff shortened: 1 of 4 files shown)


bill-roemhild:
Was logging every line of the comicinfo.xml.  Not needed.

evilhero:
We can't add all of this as one single PR - there are too many commits included that overlap and will cause issues when we merge into master. You'll have to recommit single PR's that don't make changes to the same file multiple times within the same PR. 

Alot of the later changes made are for formatting related, which honestly, we're not going to commit formatting /lint changes at this time as the whole importer module is in the process of a rewrite so almost all of these changes (formatting and otherwise) honestly will probably be moot.

I'm also still on the fence as to the inclusion of metadata within cbr files - all of the changes could have been avoided if you were to convert to cbz (since you're writing metadata to cbr, you're not worried about retaining the original as is).  While I'm not against it entirely, calling the module for rar extraction is probably easier than copying the code from the other module.

All being said we do appreciate the PR's and the help with the bugs obviously - if you resubmit the PR's as stated above we can better merge them in in sequence and deal with them as individual changes as opposed to one enormous PR.

bill-roemhild:
Thanks, I'm no longer going to be submitting changes.  I was moving way to quick for my own good and will now just be updating my fork only.  

Thanks for a starting point.

PR #1360, opened 2023-05-30
Fall back to webpage for newer ComicTagger versions

bbtufty:
Newer comictagger versions don't add an [Issue ID XXXXXX] to the notes part of metadata. This adds a fallback if not found, to pull the issue ID instead from the webpage (if possible)


evilhero:
Thanks for the PR!

My only concern is what I mentioned in the issue, is that you should make sure that the url being parsed is from CV and not from another website or else issueid will be something completely unrelated to CV.

bbtufty:
Edited, there's now that check in there

evilhero (approved):
Good to go!

PR #1362, opened 2023-05-31
FIX #1361: Use shutil.move to support cross-device renames

### LINKED ISSUE #1361: Cannot change the directory of a comic across devices
**Describe the bug**
When using multiple mounts in Docker (or multiple devices in bare metal installation) comics cannot be moved from one device to another. For example I have two separate block devices mounted at `/comics` and `/comics2`; `/comics` is the default location, but I want this particular comic to be on `/comics2`. If I change the Directory of the comic from `/comics/The Sandman V2 (1989)` -> `/comics2/The Sandman V2 (1989)`  it will fail. 

For whatever reason the error message is not logged in the log files. For lack for more information I have included the traceback. 

```
  File "/app/mylar3/mylar/webserve.py", line 6947, in comic_config
    os.rename(orig_location, com_location)
OSError: [Errno 18] Cross-device link: '/comics/The Sandman V2 (1989)' -> '/comics2/The Sandman V2 (1989)'
```

According to the [`os.rename`](https://docs.python.org/3/library/os.html#os.rename) documentation `os.rename` does not support cross-device renames and `shutil.move` should be used instead. 

 - OS: linuxserver/mylar3 docker image
 - Branch / commit: v0.7.2 (master eca80ef644760c3d7109800e7f6b247c39e648ce), Python 3.10.11

beardypig:
`os.rename` does not support cross-device renames and `shutil.move` should be used instead to support cross-device renames. 

Relevant documentation: https://docs.python.org/3/library/os.html#os.rename

### DIFF
--- a/mylar/webserve.py
+++ b/mylar/webserve.py
@@ -6944,7 +6944,7 @@ class WebInterface(object):
                             logger.fdebug('making directory: %s' % os.path.split(com_location)[0])
                             os.mkdir(os.path.split(com_location)[0])
                         logger.info('Renaming directory: %s --> %s' % (orig_location,com_location))
-                        os.rename(orig_location, com_location)
+                        shutil.move(orig_location, com_location)
                     except Exception as e:
                         if 'No such file or directory' in e:
                             checkdirectory = filechecker.validateAndCreateDirectory(com_location, True)


barbequesauce:
Simple enough... but we don't commit to master. Please resubmit to the dev branch.

PR #1368, opened 2023-06-29
Kubernetes detection

lavahot:
Hi, this is adding some Kubernetes detection. Really, it should point to the same detection because this is not DRY, but whatev.

Inspired by #1241

### DIFF
--- a/mylar/config.py
+++ b/mylar/config.py
@@ -858,7 +858,7 @@ class Config(object):
                             if config.has_section(section):
                                 config.remove_option(section, ini_key)
                             if len(dict(config.items(section))) == 0:
-                                config.remove_section(section) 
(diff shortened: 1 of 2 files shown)


evilhero:
Firstly, thanks for the PR! 

It looks fine on this side, but we can't merge it as it contains the actual python3-dev merge as one of the commits of this PR. 

The PR can only contain your changes - so you'd have to do a git pull first on the dev branch, then change the branch to a new one, make your changes and then PR against that (think I said that right).

We'll be able to merge it once that's done.

Closing this so that a clean PR can be done (and we don't get confused)

lavahot:
Curse you, Github editor! Okay, I'll figure that one out. BRB.

PR #1373, opened 2023-07-10
getcomics: correct logging call

OddBloke:
When hitting a `ConnectionError`, I observed the following in my log:

```
During handling of the above exception, another exception occurred:

  File "/app/mylar3/mylar/logger.py", line 337, in new_run
    old_run(*args, **kwargs)
  File "/usr/lib/python3.10/threading.py", line 953, in run
  File "/app/mylar3/mylar/helpers.py", line 3241, in ddl_downloader
    ddzstat = ddz.downloadit(item['id'], item['link'], item['mainlink'], item['resume'], item['issueid'], remote_filesize)
  File "/app/mylar3/mylar/getcomics.py", line 923, in downloadit
    logger.error('[ERROR] download has timed out due to inactivity...' % e)
TypeError: not all arguments converted during string formatting
```

This caused the DDL lock to never be released, requiring me to restart Mylar3 for any direct downloads to occur.

This commit corrects that `TypeError`, allowing exception handling to continue as it should (and so releasing the lock).

### DIFF
--- a/mylar/getcomics.py
+++ b/mylar/getcomics.py
@@ -920,7 +920,7 @@ class GC(object):
                                 f.flush()
 
         except requests.exceptions.Timeout as e:
-            logger.error('[ERROR] download has timed out due to inactivity...' % e)
+            logger.error('[ERROR] download has timed out due to inactivity...: %s', e)
             mylar.DDL_LOCK = False
             return {"success": False, "filename": filename, "path": None}
 


evilhero:
Oh ffs - yeah that was a typo on my part dating back months ago that was never caught.

Good catch and thanks for the PR!

evilhero (approved):

evilhero:
Sorry I didn't catch this cause the github app didn't show me all the changes. This isn't about the PR, but all the extra stuff it's trying to bring in with it. 

I'm guessing because you changed the branch after the PR that it's trying to bring in the dev merges all at once in addition to your PR. This throws off the merges later when we rebase and causes problems for us.

If you can do your commit against the python3-dev branch (prior to creating the PR, do a git pull, then fork, then commit kinda thing), and then submit a new PR again, we'll be able to merge it no problem.

Sorry for the inconvenience, it may seem minor but it can literally take hrs to manually go thru each commit during a rebase if

OddBloke:
Thanks for Mylar, and thanks for the speedy review! I did rebase my commit onto `python3-dev` (_after_ I switched to the correct base branch), and I'm only seeing [a single commit](https://github.com/OddBloke/mylar3/commit/fe13cf371abab1fb24430c39d1f6196673f25f41) in https://github.com/mylar3/mylar3/pull/1373/commits (whose parent looks correct: 4b9c557). Is it possible you were seeing something that was generated before I did that rebase?

evilhero:
I'm seeing the PR go against the master branch, then you did a branch change and then committed the branch change in addition to the commit - so it looks like it's picking up on the initial branch change which is in there when I go to do the rebase. Doing the PR against the dev branch initially and not committing changing branches in the PR branch (not sure if that was intentional or not) would be enough to line things up I'd imagine. 

So I'm not sure what's going on, but we generally err on the side of caution when we rebase in stuff, as when we merge up to master, repeated rebases of the same items will cause the merge to become a manual intensive task.

OddBloke:
I'll close this and open a new PR as requested!

PR #1382, opened 2023-08-01
Add SOCKS proxy support

### LINKED ISSUE #1282: FEATURE REQUEST - Proxy option for Mylar
**Describe the request**
Similar to Sonarr/Radarr etc where I'm able to have Mylar operate through a proxy. My personal use case is on unRAID with a docker container stack where one is the VPN container with a Privoxy port.

While indexers are all proxied in Prowlarr, as is Prowlarr itself, requests are still made and viewed by Mylar. Additionally (and perhaps most importantly) DDL and the built-in support for DOG and SU pose an issue without some form of masking from ISPs in cyber-oppressive countries like Australia where I live, regardless of if the actual download from NZB itself is encrypted (which would enough fine for most countries).

**How did you come up with the idea?**
Got the idea from the implementation visible in Sonarr, Radarr, Readarr, Lidarr, Whisparr, Prowlarr, and Jackett

 - OS: unRAID

Thanks for your consideration,
Avs

NickSica:
Third times the charm, git is not my strong suit for sure! Here is a reupload of #1379, hopefully I didn't mess anything up. I'll repeat the original body here for clarity.

Should close #1282. Also there was a weird bug in the Warnings/Important Release Info probably due to the fact that the socks optional dependency for requests was included using brackets(there's probably a better way to do this perhaps just downloading PySocks separately?).

Using configuration file, set ENABLE_PROXY to true and set the addresses of the proxy. An http proxy would be set like "http://myproxy.com/" while a SOCKS5 proxy would be set like "socks5://user:[password@myproxy.com](mailto:password@myproxy.com):port".

Check out [this](https://docs.python-requests.org/en/latest/user/advanced/#socks) for more info.


evilhero (approved):
good to go - I'll deal with the config page related problem once this is merged up (should be a simple fix).

PR #1383, opened 2023-08-02
getcomics: support fetching all search results using pagination

OddBloke:
This series of commits refactors the fetching of getcomics results pages into a separate method, and implements iterating through all of the search result pages.

I've tested this locally using Fantastic Four (2018) #3. It has a year mismatch between Mylar and GC, so the query which returns it is "Fantastic Four #3" which has enough results that it falls onto the second page. With trunk, the issue isn't found; with this branch, it is.

(I've broken the refactor down into logical commits, so reviewing commit-by-commit may make your life easier!)

PR #1384, opened 2023-08-02
getcomics: support fetching all search results using pagination

OddBloke:
This series of commits refactors the fetching of getcomics results pages into a separate method, and implements iterating through all of the search result pages.

I've tested this locally using Fantastic Four (2018) #3. It has a year mismatch between Mylar and GC, so the query which returns it is "Fantastic Four #3" which has enough results that it falls onto the second page. With trunk, the issue isn't found; with this branch, it is.

(I've broken the refactor down into logical commits, so reviewing commit-by-commit may make your life easier!)

### DIFF
--- a/mylar/getcomics.py
+++ b/mylar/getcomics.py
@@ -372,9 +372,7 @@ class GC(object):
         resultlist = []
         soup = BeautifulSoup(page_html, 'html.parser')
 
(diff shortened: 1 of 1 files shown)


OddBloke:
(This started life as #1383.)

evilhero (approved):

PR #1385, opened 2023-08-04
helpers: handle absent filename in ddzstat

OddBloke:
When downloading a pack, `ddzstat['path']` is the unpacked directory and `ddzstat['filename']` is set to `None`.  This commit avoids passing `None` to `os.path.join`, avoiding an exception.


evilhero (approved):

evilhero:
Looks Good ! Thanks for the PR(s) again!

PR #1388, opened 2023-08-07
FIX: Update configuration screen to allow for requests[socks] in requirements.txt

evilhero:
Configuration screen would indicate pip failure due to requests[socks] requirement change regardless of if it was installed or not.

### DIFF
--- a/mylar/req_test.py
+++ b/mylar/req_test.py
@@ -78,7 +78,11 @@ class Req(object):
                                 p_version = str(line[p_test+2:]).strip()
                                 if p_version == '':
                                     continue
-                                self.req_list.append({'module': p_mod, 'version': p_version, 'arg': p_arg})
+                                if p_mod == 'requests[socks]':
+                                    self.req_list.append({'module': 'requests', 'version': '2.22', 'arg': '>='})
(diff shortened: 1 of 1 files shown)


mylar/req_test.py
```diff
@@ -78,7 +78,11 @@ def check_the_pip(self):
                                 p_version = str(line[p_test+2:]).strip()
                                 if p_version == '':
                                     continue
-                                self.req_list.append({'module': p_mod, 'version': p_version, 'arg': p_arg})
+                                if p_mod == 'requests[socks]':
+                                    self.req_list.append({'module': 'requests', 'version': '2.22', 'arg': '>='})
```
barbequesauce, line 82: Hardcoding version numbers? No way to reference requirements.txt current values?

barbequesauce (approved):

mylar/req_test.py
evilhero: Yeah was pretty ugly right? 

Because we're importing socks as part of requests it only holds requests version number.  And when I did on my ancient system is installed 1.7.1, so I just took a lower value in case some bloke is running older hardware than I am haha. 

The other alternative is to just add ``PySocks`` to the requirements.txt and change ``requests[socks``] back to ``requests``.

PR #1392, opened 2023-08-15
getcomics: stop searching when first matching issue is found

OddBloke:
There are two main parts to this PR:

* it refactors `search_filer`s per-issue logic into a separate method, which allows us to reuse it for both the existing "get all matches" method, as well as a new "get the first match" method
* converts `getcomics`' `perform_search_queries` to generate each result, meaning we only fetch the next page when all of the results on the current page have been processed (as well as some associated refactoring to avoid exhausting the generator immediately)

Putting these together means that Mylar will now only fetch the next GC results page if it has not found a match on the current page: this should result in fewer results pages being fetched, reducing load on GC and improving search times.

I've split this down into a (hopefully) sensible sequence of commits. Note that the refactoring lead to a mass dedent, so I strongly recommend reviewing this with `?w=1` appended to the GitHub URL or `-w` passed to `git diff`: that gives a more accurate view of 58 additions and 67 deletions.

### DIFF
--- a/mylar/findcomicfeed.py
+++ b/mylar/findcomicfeed.py
@@ -150,7 +150,6 @@ def Startit(searchName, searchIssue, searchYear, ComicVersion, IssDateFix, bookt
             return "no results"
 
     entries = []
-    mres = {}
     totNum = len(feed.entries)
     keyPair = []
     regList = []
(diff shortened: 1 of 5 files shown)


OddBloke:
(I've just pushed up a commit to restore honouring of PACK_PRIORITY.)

evilhero (approved):

evilhero:
Looks good - tested it on a few GC and newznab entries and everything flew threw just fine as expected.

Thanks for the PR!

OddBloke:
Thanks for the review and the merge!

PR #1393, opened 2023-08-15
getcomics: handle literal-HTML descriptions which occasionally occur

OddBloke:
I've run into issues on GC which have the short description HTML escaped
and so appearing as normal text: this means that it isn't queryable
using `.find`.  This commit detects that case and creates a new
`BeautifulSoup` object for the escaped HTML, and then searches it.

I've seen this fix searching for Batman (2016) issue 12, as it has such a
description.

(This is another PR with a dedent, so I recommend reviewing this with `?w=1` appended to the GitHub URL or `-w` passed to `git diff`.)

### DIFF
--- a/mylar/getcomics.py
+++ b/mylar/getcomics.py
@@ -462,10 +462,18 @@ class GC(object):
                 ):
                     continue
 
-            option_find = f.find("p", {"style": "text-align: center;"})
+            needle_style = "text-align: center;"
+            option_find = f.find("p", {"style": needle_style})
             i = 0
             if option_find is None:
-                continue
+                # Some search results have the "option_find" HTML as escaped
+                # text instead of actual HTML: try to salvage a option_find in
+                # that case
+                excerpt = f.find("p", class_="post-excerpt")
+                if excerpt is not None and needle_style in excerpt.text:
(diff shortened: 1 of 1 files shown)


mylar3:
I believe this is not uptodate anymore due to the previous PR merge of yours - same file changes the diff and it's showing that if I rebase this into the python3-dev branch it will restore all the changes you did in your previous PR, which isn't what we want.

I haven't figured out the github status actions to denote this (as of yet), but I think you probably need to close this, update your local branch and then do the PR against that again...

Pretty sure because these are all touching the same file repeatedly that each of the other PRs in the queue will be in the same boat as we merge/rebase things up.

OddBloke:
> I haven't figured out the github status actions to denote this (as of yet), but I think you probably need to close this, update your local branch and then do the PR against that again...

I shouldn't need to close/open the PR: they follow force-pushes. As you can see just above this comment (https://github.com/mylar3/mylar3/pull/1393#event-10134983870), I've just force-pushed this, so it's now just two commits on top of `python3-dev`.

> Pretty sure because these are all touching the same file repeatedly that each of the other PRs in the queue will be in the same boat as we merge/rebase things up.

Yep, I'll keep on top of updating and force-pushing them as needed!

PR #1397, opened 2023-08-18
FIX: queues were not firing off due to a logging error that aborted the startup queue process

evilhero:
resulting in no queues being started


mylar3 (approved):

OddBloke:
Oops, sorry about this! I've got a running dev instance set up now, so I should catch errors like this in future.

Issue #1403, opened 2023-08-21
Error when mylar searches for issues which are to be downloaded.


jindalpratik:
**Describe the bug**
Mylar blows up when trying to search for any comic.

**CarePackage**
https://quickshare.samsungcloud.com/axUaV8PwmKVI

**Environment (please complete the following information):**
 - OS: Windows 11
 - Branch / commit:  aa00fab3e574475d0452a3f9b2164a18b6b206f9 (python3-dev)

OddBloke:
This is the traceback:
```pytb
21-Aug-2023 14:44:04 - ERROR   :: mylar.searchforissue.2568 : SEARCH-QUEUE : Traceback (most recent call last):
  File "C:\Users\dapte\mylar3\mylar\search.py", line 2496, in searchforissue
    foundNZB, prov = search_init(
                     ^^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 525, in search_init
    findit = search_the_matrix(scarios)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 4209, in search_the_matrix
    return NZB_SEARCH(
           ^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 1128, in NZB_SEARCH
    if len(bb['entries']) > 0:
           ~~^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
Traceback (most recent call last):
  File "C:\Users\dapte\mylar3\mylar\search.py", line 2496, in searchforissue
    foundNZB, prov = search_init(
                     ^^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 525, in search_init
    findit = search_the_matrix(scarios)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 4209, in search_the_matrix
    return NZB_SEARCH(
           ^^^^^^^^^^^
  File "C:\Users\dapte\mylar3\mylar\search.py", line 1128, in NZB_SEARCH
    if len(bb['entries']) > 0:
           ~~^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

OddBloke:
I've reproduced the error locally, and confirmed that #1404 fixes it for me.

PR #1404, opened 2023-08-21
rsscheck: undo erroneous changes to nzbdbsearch's return type

OddBloke:
In dab1bf372fba803c58516bd78d1ef94e721bd42b I simplified the return types of functions whose return values are passed to `search_check.checker`.  `nzbdbsearch`'s return value is used in a different way, and so I should never have changed it.  Apologies!

Fixes: #1403


OddBloke:
I have reproduced the error locally, and confirmed that this fix resolves it.

evilhero (approved):

PR #1410, opened 2023-08-29
Expose current queue status at /prometheus_metrics for Prometheus monitoring

OddBloke:
Also includes some refactoring to the queue view in Manage, as it's the same data being presented.

This allows people with a Prometheus instance to scrape the current queue status. Here's an example of the queue sizes when downloading 9 issues (from 2 series, hence the staggered initial increase) from GC:

prometheus

It also exposes "alive" and "started" metrics which should enable people to add alerting for queues dying, if desired.

### DIFF
--- a/data/interfaces/default/manage.html
+++ b/data/interfaces/default/manage.html
@@ -286,12 +286,12 @@
                            </tr>
                         </thead>
                         <tbody>
-                           %for q_name, (is_up, length) in queues.items():
+                           %for q in queues:
                               <%
-                                    if is_up:
+                                    if q.is_alive:
                                         text = "Up"
                                         grade = '#7AEE52'
-                                    elif is_up is not None:
+                                    elif q.is_alive is not None:
                                         text = "Down"
                                         grade = '#EC7564'
                                     else:
@@ -299,8 +299,8 @@
                                         grade = '#52A9EE'
                                %>
                               <tr>
-                                <td style="width: 50px;text-align: center;">${q_name}</td>
-                                <td style="width: 50px;text-align: center;">${length}</td>
+                                <td style="width: 50px;text-align: center;">${q.name}</td>
+                                <td style="width: 50px;text-align: center;">${q.size}</td>
                                 <td style="width: 50px;text-align: center;color:${grade}">${text}</td>
                               </tr>
                            %endfor
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -34,6 +34,7 @@ import hashlib
 import gzip
 import os, errno
 import urllib
+from collections import namedtuple
 from urllib.parse import urljoin
 from io import StringIO
 from apscheduler.triggers.interval import IntervalTrigger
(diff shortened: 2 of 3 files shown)


OddBloke:
(Switched base: the commits were already based on `python3-dev`, so there should be no problems landing them.)

mylar3:
I'm still getting errors on my side when I run the rebase into python3-dev primarily due to the branch change after the commits where done against the master branch it looks like (it might have to do with us doing a rebase when we go up to master along with the enforcement of trying to keep a linear history and the initial head being set to master).

I know it's a pain, but could you do a clean PR against the python3-dev branch and then I can merge this into dev? 

I'll merge in your other PR since it seems to be clean in the meantime.

OddBloke:
> I'm still getting errors on my side when I run the rebase into python3-dev primarily due to the branch change after the commits where done against the master branch it looks like (it might have to do with us doing a rebase when we go up to master along with the enforcement of trying to keep a linear history and the initial head being set to master).
>
> I know it's a pain, but could you do a clean PR against the python3-dev branch and then I can merge this into dev?

I'd like to figure this out, if we can, to avoid having to repeat this in future (and because I'm curious how we're ending up here!). These commits were always based on a `python3-dev` commit, and I don't _think_ the base branch in GitHub is used, before merging, for anything other than diff generation/conflict detection.

I've just rebased my commits onto the latest `python3-dev`, and this is now what the history looks like:

```
* e0971234 - (HEAD -> prometheus_metrics, origin/prometheus_metrics) webserve: introduce /prometheus_metrics with queue metrics (3 minutes ago) <Daniel Watkins>
* 6345805f - helpers: convert queue_info to a generator (3 minutes ago) <Daniel Watkins>
* 1cc1d09e - helpers: convert queue_info to return a namedtuple (3 minutes ago) <Daniel Watkins>
* 37947705 - webserve: refactor queue information generation to helpers (3 minutes ago) <Daniel Watkins>
* 10902054 - (upstream/python3-dev) FIX: DDL failing to search/download due to CF/FS problems (16 hours ago) <evilhero>
```

As you can see, my commits are all sitting on top of the most recent `python3-dev` commit, so they should fast-forward merge/rebase with no problems. Do you still see issues now?

> I'll merge in your other PR since it seems to be clean in the meantime.

Thanks!

evilhero:
Sorry about the delay in this - life happened (along with a bit of Starfield in there).

This looks fine now - I use gitkraken sometimes to check out the linear workflow and to see how far off items are. I unfortunately didn't take a snapshot of the previous PR, but I've attached what I'm seeing now:

yqtkf5l0

What I saw previously was that the python3-dev branch was placed after your previous PR, and the point where it forked deviated from the master branch instead of python3-dev. 

But it's all in line now, so I don't have any reason to have any concerns at this point. The reason why I'm overly cautious on these things is because there was a few instances awhile go where we accepted PR's that had branch changes / previous commited PR's and I had to basically recreate the entire branch in order to get rid of the problematic entries, which wasn't fun cause I was having to cherrypick thru every commit prior, removing the bad ones and having to verify the commits weren't actually rebasing to a prior commit).

evilhero (approved):

PR #1417, opened 2023-09-08
manage: correct wording about worker repair

OddBloke:
I misunderstood what the DDL "restart" button does: it requeues items, but it does not restart the worker itself.


evilhero:
Based on our previous discussions, this PR needs to be rebased / force-pushed so it's current and doesn't lose the previous changes to manage.html:

k41wyi1w

Will merge up once it's current. Cheers!

evilhero (approved):

PR #1435, opened 2023-10-28
fix(search): Skip searching for an issue if it's not 'Wanted'

morsok:
This will prevent the search of issues that are marked snatched because mylar found a pack for them.


morsok:
Yeah I don't even remember the use case now, closing.

PR #1460, opened 2023-12-08
FIX: pack downloads were broken due to previous PR variable change

evilhero:
Also, when downloading packs:

on the Manage DDL page, will now show the full filename of the pack instead of ``series #issuenumber`` (ie. ``Superhero # 1``)


barbequesauce (approved):
Ooopsie

PR #1496, opened 2024-01-27
post process for manually grabbed one-off storyarc issues

nellistc:
Adds new option (oneoff2arcdir) to allow manually grabbed one-off storyarc issues to be copied to the configured storyarc directory without copying watchlisted issues. Uses existing code to do the copying/moving/linking, only changes the option used to trigger the file action.

Option added to storyarc pages. Not really sure about the wording. Thought about:

[x] [copy] one-off issues to StoryArc directory
 -- [x] include watchedlisted issues

I didn't want to change the meaning of the existing copy2arcdir option.

### DIFF
--- a/data/interfaces/default/storyarc.html
+++ b/data/interfaces/default/storyarc.html
@@ -83,6 +83,9 @@
                                               <option value=${x} ${outputselect}>${x}</option>
                                          %endfor
                                         </select><label> watchlisted issues to <span id="arcopts"></span> Directory</label>
+                           <div>
+                               <input type="checkbox" style="vertical-align: middle; margin: 3px; margin-top: -1px;" name="oneoff2arcdir" id="oneoff2arcdir" value="1" ${checked(mylar.CONFIG.ONEOFF2ARCDIR)} /><label>Process one-off issues</label>
+                           </div>
                              <div id="arc_fileops_options">
                                  <input type="checkbox" style="vertical-align: middle; margin: 3px; margin-top: -1px;" name="arc_fileops_softlink_relative" id="arc_fileops_softlink_relative" value="1" ${checked(mylar.CONFIG.ARC_FILEOPS_SOFTLINK_RELATIVE)} /><label>Create relative softlink <span style="font-size:10px;color:#999;">(default: absolute)</span></label>
                                  </div>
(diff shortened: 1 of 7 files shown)


evilhero:
Firstly thanks for the PR!

In reading thru it, I'm abit at a loss as to the reason behind it (it is really late here in my defense, so I may be overlooking something obvious).  

I think where I'm confused abit is that if it's detected as a one-off issue (meaning it's not on the watchlist), it wouldn't be considered a watchlisted item anyways. The post-processing workflow is also on a per issue basis - it'll run thru the module one issue at a time in sequence that it matches to (again, maybe I'm thinking you think it's doing batch processing instead of a FIFO-type of queue).

Removing aspects in the post-processor for the copy2arc option maybe is where I'm also getting confused, as it seems that you are maybe just changing the variable name to get it to post-process in your particular workflow? 

If the copy2arc option isn't enabled and it's a one-off issue, the intent is that it is moved to the storyarc directory. The only time when things get dropped into the grabbag_directory is when the storyarc directory option isn't enabled, and it encounters a non-watchlisted storyarc issue as it has to have somewhere to put it (probably not what you're doing or intending).

The only other thing that bears mention is that we never merge stuff directly into master, it's always against the python3-dev branch. So if we were to merge this (or even continue the convo), it should be under a PR against an up-to-date python3-dev branch (you'd have to do a clean PR against it).

Sorry for the ramble, late here but I do think maybe a bit more detailed explanation as to why it's needed or perhaps your use case for it would help immensely in us understanding things.

nellistc:
> In reading thru it, I'm abit at a loss as to the reason behind it (it is really late here in my defense, so I may be overlooking something obvious).

Reading your comment, it could be more of a bug fix and may not be necessary.

> If the copy2arc option isn't enabled and it's a one-off issue, the intent is that it is moved to the storyarc directory. The only time when things get dropped into the grabbag_directory is when the storyarc directory option isn't enabled, and it encounters a non-watchlisted storyarc issue as it has to have somewhere to put it (probably not what you're doing or intending).

In my testing, if copy2arcdir isn’t enabled, no issues get copied to the storyarc directory. That’s what this PR attempts to fix. If it is enabled, all issues from the story arc (including issues from watchedlisted series) are copied to the storyarc directory.

> The only other thing that bears mention is that we never merge stuff directly into master, it's always against the python3-dev branch. So if we were to merge this (or even continue the convo), it should be under a PR against an up-to-date python3-dev branch (you'd have to do a clean PR against it).

Yep, my mistake - I’ll redo if this progresses.

Thanks

nellistc:
no longer required

PR #1506, opened 2024-02-08
IMP: Detect and handle CV reusing existing ComicIDs

evilhero:
Will check existing information in db with what CV is telling it is supposed to be. It will now do 4 comparison checks - if > 50% of the checks fail, it will assume it's been changed and indicate as such. It will then:
- automatically pause the series
- flag it as ``cv_removed`` color coded on the main watchlist page
- when loading the series detail page for the given series, it will prompt with an explanation and 2 choices:
   - delete: delete the series from the watchlist (also an option to delete the series from the disk)
   - keep: will retain the series as is with no further information being retrieved from CV

*Note: this also adds a color-coded legend to the bottom of the main watchlist/index page, along with color-coded rows now for  status of : ``cv_removed``, ``loading``, ``paused``, and ``failed/error``.

### DIFF
--- a/data/interfaces/carbon/css/style.css
+++ b/data/interfaces/carbon/css/style.css
@@ -1376,7 +1376,6 @@ div#artistheader h2 a {
   vertical-align: middle;
 }
 #series_table {
-  background-color: #FFF;
   padding: 20px;
   width: 960px !important;
 }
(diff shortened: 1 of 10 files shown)


barbequesauce (approved):
not throwing errors for me on load or one searching/adding a known reused series.....

PR #1512, opened 2024-02-17
FIX: dashboard search missing volumes when compared to ComicVine

Fmstrat:
When searching from the dashboard (logged in home page) Mylar uses the `/volumes` API endpoint with a `name:` filter. This results in an over-filtered result set that does not match with ComicVine's site.

Example:
```
https://comicvine.gamespot.com/search/?i=volume&q=soulfire%20test
```
Returns additional results when compared to the currently used Mylar URL:
```
https://comicvine.gamespot.com/api/volumes?api_key=<YOURKEY>&filter=name:soulfire,name:test&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=xml&sort=date_last_updated:desc&offset=0
```
By adjusting the search to use the `/search` endpoint, with a `resources` filter set to `volumes`, we get the same format XML result, but it includes the search results that ComicVine would intend:
```
https://comicvine.gamespot.com/api/search?api_key=<YOURKEY>&query=soulfire test&resources=volume&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=json&sort=date_last_updated:desc&offset=0
```
This PR only adjusts this for `search_type` of `None`, so it does not impact other areas of the application at this time.

### DIFF
--- a/mylar/mb.py
+++ b/mylar/mb.py
@@ -48,18 +48,21 @@ if platform.python_version() == '2.7.6':
 
 def pullsearch(comicapi, comicquery, offset, search_type):
 
-    cnt = 1
-    for x in comicquery:
-       if cnt == 1:
-           filterline = '%s' % x
-       else:
-           filterline+= ',name:%s' % x
-       cnt+=1
-
-    PULLURL = mylar.CVURL + str(search_type) + 's?api_key=' + str(comicapi) + '&filter=name:' + filterline + '&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=xml&sort=date_last_updated:desc&offset=' + str(offset) # 2012/22/02 - CVAPI flipped back to offset instead of page
+    pull_url = mylar.CVURL
+    if search_type is 'volume':
+        pull_url = pull_url + 'search?api_key=' + str(comicapi) + '&query=name:' + comicquery + '&resources=' + str(search_type) + '&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=xml&sort=date_last_updated:desc&offset=' + str(offset) # 2012/22/02 - CVAPI flipped back to offset instead of page
+    else:
+        cnt = 1
+        for x in comicquery:
+            if cnt == 1:
+                filterline = '%s' % x
+            else:
+                filterline+= ',name:%s' % x
+            cnt+=1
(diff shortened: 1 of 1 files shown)


evilhero:
This is the expected result. We use the filter so it acts like an ``AND`` btwn each word of the query. Yours does the exact opposite and uses ``OR`` which will balloon the results to include items that have no need to be included (ie. ``test`` or ``test dirt`` both of which aren't what was queried for). Your query for example if using ``Batman Off-world`` will return over 3000 hits (10 api hits since we limit results to the first1000 regardless) , the current iteration returns 1 correct result with 1 api hit. 

Since CV has gone and made using their api even more difficult to use due to api hit restrictions, implementing this would be a very bad idea as you'll get an extreme amount of hits that would be avoided under the current implementation, thereby causing excessive api hits. 

While we thank you for the time it took to do the PR and to submit it, we won't be merging this,
Outside of the fact that it's against the wrong branch (should always be python3-dev), it's just not a valid approach to take given CV's current api regime.

Fmstrat:
@evilhero 

Ahh good points. Before closing, may we discuss potential revisions? Pulling `test` out of the equation, the specific use case that led me here was this:

- Looking for `Aspen Soulfire Revelations`, as it is commonly called when speaking about the Soulfire series. The formal name for this TPB is `Aspen Universe: Revelations`, though a user, such as myself, would not have known this.
- The search then uses:
  - `https://comicvine.gamespot.com/api/volumes?api_key=<KEY>&filter=name:aspen,name:soulfire,name:revelations&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=json&sort=date_last_updated:desc&offset=0`
- "0 results returned" (due to the `OR`)
- Have to go to ComicVine to figure this out

A handier experience might be:
- ... the above ...
- "0 results found. (link)Expand your search?(/link)"
- Switch to the `OR` (maybe limit to top 10?):
  - `https://comicvine.gamespot.com/api/search?api_key=<KEY>&query=aspen soulfire revelations&resources=volume&field_list=id,name,start_year,site_detail_url,count_of_issues,image,publisher,deck,description,first_issue,last_issue&format=json&sort=date_last_updated:desc&offset=0`
- Top result is the match

This way it's not the default behavior, and would only allocate the additional API hits to those rare occurrences someone needs it.

Also apologies on the wrong branch, I didn't see a `CONTRIBUTING` doc, but guess I could have joined on IRC fist.

PR #1539, opened 2024-03-29
Fix not defined issueline

halkeye:
NameError: name 'issueline' is not defined

fixes https://github.com/mylar3/mylar3/issues/1180


evilhero:
We thank you firstly for the PR. However we cannot accept it as is, due to the force-push from the master branch being present. 

Please do a clean PR against the python3-dev branch instead of forking the master and doing a forced push against the dev. (ie. Checkout the dev branch, git pull it, then fork and do your changes in the fork of the dev branch. PR against that fork and you're good to go). 

The only thing we should see listed for commits is the changes you're doing, not changes we have already pushed.

I'll close this and then you can do a new PR as mentioned above and then we can merge it up no problem.

PR #1540, opened 2024-03-29
Fix not defined issueline

halkeye:
NameError: name 'issueline' is not defined

fixes https://github.com/mylar3/mylar3/issues/1180


evilhero:
Merged into python3-dev, 

Much appreciation for the PR and apologies for the massive delay - easier to manually push than request a redo, so gave thanks in PR #1600

PR #1550, opened 2024-06-28
[FIX] fix error if comicinfo is None in DDL queue

Bart274:
This MR solves this error:

```
During handling of the above exception, another exception occurred:

  File "C:\Users\de_va\mylar3\mylar\logger.py", line 337, in new_run
    old_run(*args, **kwargs)
  File "C:\python310\lib\threading.py", line 953, in run
  File "C:\Users\de_va\mylar3\mylar\helpers.py", line 3428, in ddl_downloader
    ggc.parse_downloadresults(item['id'], item['mainlink'], item['comicinfo'], item['packinfo'], link_type_failure[item['id']])
  File "C:\Users\de_va\mylar3\mylar\getcomics.py", line 1065, in parse_downloadresults
    self.issueid = comicinfo[0]['IssueID']
TypeError: 'NoneType' object is not subscriptable
```

### DIFF
--- a/mylar/PostProcessor.py
+++ b/mylar/PostProcessor.py
@@ -440,11 +440,11 @@ class PostProcessor(object):
                         return self.queue.put(self.valreturn)
                     logger.info('I have located %s files that I should be able to post-process. Continuing...' % filelist['comiccount'])
                 else:
-                    if all([self.comicid is None, '_' not in self.issueid]):
+                    if all([self.comicid is None, self.issueid is not None and '_' not in self.issueid]):
                          cid = myDB.selectone('SELECT ComicID FROM issues where IssueID=?', [str(self.issueid)]).fetchone()
                          self.comicid = cid[0]
                     else:
-                         if '_' in self.issueid:
+                         if self.issueid is not None and '_' in self.issueid:
                              logger.fdebug('Story Arc post-processing request detected.')
                              self.issuearcid = self.issueid
                              self.issueid = None
--- a/mylar/getcomics.py
+++ b/mylar/getcomics.py
@@ -1061,11 +1061,11 @@ class GC(object):
                 logger.warn('[GC-Site-Unknown] Unknown site detected...%s' % lt_site)
                 link_type = 'Unknown'
 
-            if self.issueid is None:
+            if self.issueid is None and comicinfo is not None:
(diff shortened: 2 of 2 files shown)


mylar/PostProcessor.py
```diff
@@ -440,11 +440,11 @@ def Process(self):
                         return self.queue.put(self.valreturn)
                     logger.info('I have located %s files that I should be able to post-process. Continuing...' % filelist['comiccount'])
                 else:
-                    if all([self.comicid is None, '_' not in self.issueid]):
+                    if all([self.comicid is None, self.issueid is not None and '_' not in self.issueid]):
```
cuthbeorht, line 443: Would it be clearer to have a variable encompass all the conditional logic amking it easier to read?

```suggestion
                    _is_comic_info_none = all([self.comicid is None, self.issueid is not None and '_' not in self.issueid])
                    if _is_comic_info_none:
```

cuthbeorht:
I think the [documentation](https://mylarcomics.com/docs/contributing#pull-requests) specifies to create a PR off of `python3-dev` and not the master branch.

cuthbeorht:
Good thinking on checking that condition! 💪

barbequesauce:
Thanks - as @cuthbeorht mentions, while we appreciate the PR we prefer to run against the dev branch for a while to shake things out. Please resubmit against dev…

evilhero:
My main concern with this PR is that comicinfo should never be empty, so while this PR does address the problem, it doesn't actually fix it as without the comicinfo being present, then it most likely came from a failed retry or something similar that would cause empty values to propagate.

cuthbeorht:
> My main concern with this PR is that comicinfo should never be empty, so while this PR does address the problem, it doesn't actually fix it as without the comicinfo being present, then it most likely came from a failed retry or something similar that would cause empty values to propagate.

Good point. I was missing the context. Is there a process for tying a PR to an issue?

francispeixoto:
> My main concern with this PR is that comicinfo should never be empty, so while this PR does address the problem, it doesn't actually fix it as without the comicinfo being present, then it most likely came from a failed retry or something similar that would cause empty values to propagate.

I get that this is most likely a bandaid over the real issue, but it would most likely allow the DDL flow to start up again until a permanent fix is found, no?

PR #1552, opened 2024-07-03
Use configured http_proxy/https_proxy with pixeldrain

raph521:
The http_proxy and https_proxy config options were only used in getcomics.py for direct GC downloads.  Now, the pixeldrain downloader supports downloading through those http proxies if configured.


evilhero:
Merged into python3-dev, 

Much appreciation for the PR and apologies for the massive delay - easier to manually push than request a redo, so gave thanks in the PR #1600

Issue #1581, opened 2024-12-17
Ascii Fractions in Issue Numbers Handled Incorrectly


falo2k:
Following on from a report in #support on discord, where a user was finding issue 1½ of The Avengers (1963) was getting renamed as 000.5.  There is no 000.5 for that series.

The cause of the problem is that the issue number is stored in CV as 1½ using the ascii character of the fraction.  In the post processor, this line assumes that the presence of an ascii fraction will only ever be "0.fraction" as a decimal.  I don't have any examples of comics with N 1/4 or N 3/4 where N > 0, but they would be dealt with in the same way, losing the whole number part.  

https://github.com/mylar3/mylar3/blob/ca4a3736865db0f11ed150b3768665e983c4118b/mylar/PostProcessor.py#L2784

falo2k:
Fixed for fractions in https://github.com/mylar3/mylar3/commit/fb447cd1565cc1d7bf0263521bf85eba98d5931a

PR #1582, opened 2024-12-18
Ascii fraction issue whole number fix

falo2k:
Fix for https://github.com/mylar3/mylar3/issues/1581

Happy to go back to the drawing board on this one, or take a second pass as a future PR.  I've re-used the same approach to renaming for the 1/2, 1/4, 3/4 in the renaming helper as in the post processor, but left the slightly different approaches to infinity (and accented a?) alone for now.  Could probably do with a single source of truth for converting issue numbers to formatted strings, large padded decimals, etc. as a common helper function for all of these at some point.  Maybe worth doing after this alongside a look at any other open issues for weird corner case numbering.


falo2k:
Withdrawing this as will be replaced by wider refactor.

Issue #1525, opened 2024-03-09
Bulk Metadata actions for issues inside series


DieselTech:
**Describe the request**
Currently you either metatag the entire series all over again even if you already have 90% of them tagged and use a LOT of API calls, or you have to click the metatag issue button 1-by-1. 

**How did you come up with the idea?**
When adding some issues to series it became a chore to try and get metadata for them inside mylar. If a series already has 100 issues and I went and added 20, it's too many to sit and click manual metatag on each issue. 

**Additional information**
screw CV api limits

grosejon:
This came up in the Discord recently. When you have a series with over 200 issues, you will hit your API limit before getting past issue 200. If we could select a certain number of issues and then tag just those, it would be excellent. In addition, having the ability to select one issue check box, then hold Shift and click another to select all issues between the two would be an excellent addition!

PR #1589, opened 2025-01-04
Tagging Selected Issues

falo2k:
Added a button to enable meta-tagging of a set of selected issues on a series page.

I've also added a new config variable to enable a check on trying to submit a single request for tagging for any number of issues over a threshold.  This is to try to defend against CV API limits a bit.  I've defaulted it to on, and a 200 limit.  This will prevent the new feature, and the series retag, from submitting a request for >200 tagging.  It will report an error on screen in this case.

It won't manage cumulative tracking if you do it in batches.  It also won't predict failure for the collected impact of multiple series being tagged at once from the Manage->Manage Comics tagging, but would stop any individual series within those selected which have >threshold downloaded issues as they go through the same route as the manual metatagging button.

I haven't exposed the setting in the UI as I don't think it's necessary to futz with much unless there are significant changes with CV, but happy to debate whether it should be enabled by default or not.

Feel free to squash the commits as well - I accidentally introduced a bug in the first one that I fixed in the third ...

Would close: https://github.com/mylar3/mylar3/issues/1525


falo2k:
Closing to rebase because of conflicting changes.

PR #1594, opened 2025-01-15
Fix: Support for .DEATHS issue numbering (Spider-Man 65.Deaths)

falo2k:
Had a stab at supporting the new Marvel issue naming funjoytime.  Have tested against download (GC), renaming, and manual post processing.  

The choice of making it allcaps is because of the usage of ISSUE_EXCEPTIONS that expect entries to be allcaps, and I wanted to avoid filenames changing (case) depending on what operation was occurring.

The reason for having an explicit path in the post processor rather than relying on ISSUE_EXCEPTIONS was to maintain consistency with how other dot suffix naming is handled.  If you don't do that, it introduces a bug further down because of the exceptionmatch loop (reverse order wins here, in other parts of the code a break is used so first entry wins) and you get that situation where something gets renamed into archive status.  It's a toss up on whether to tweak that potential bug, or do it this way, and I thought this way is better for the file naming outcome.


falo2k:
As discussed, will realign to latest python3-dev and resubmit.

PR #1607, opened 2025-01-24
Fix: Misc typos

falo2k:
Noticed in passing while doing other changes, isolating them to a standalone PR.  One typo in "Exception", one reference to a variable that isn't declared (and should be "cid" within the loop).


evilhero (approved):

evilhero:
Cheers!

PR #1610, opened 2025-01-25
Adding button to allow tagging of selected issues in series

falo2k:
Added a button to enable meta-tagging of a set of selected issues on a series page.

I've also added a new config variable to enable a check on trying to submit a single request for tagging for any number of issues over a threshold. This is to try to defend against CV API limits a bit. I've defaulted it to on, and a 200 limit. This will prevent the new feature, and the series retag, from submitting a request for >200 tagging. It will report an error on screen in this case.

It won't manage cumulative tracking if you do it in batches. It also won't predict failure for the collected impact of multiple series being tagged at once from the Manage->Manage Comics tagging, but would stop any individual series within those selected which have >threshold downloaded issues as they go through the same route as the manual metatagging button.

I haven't exposed the setting in the UI as I don't think it's necessary to futz with much unless there are significant changes with CV, but happy to debate whether it should be enabled by default or not.

### DIFF
--- a/data/interfaces/default/comicdetails_update.html
+++ b/data/interfaces/default/comicdetails_update.html
@@ -488,6 +488,16 @@
                         %endif
                 </select>
                 <input type="hidden" value="Go">
+                %if any([mylar.CONFIG.ENABLE_META, mylar.CONFIG.CBR2CBZ_ONLY]):
+                    <div style="padding-left: 10px; float:right">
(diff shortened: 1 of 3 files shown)


suineg:
You could expose it to the config.ini but agree it might be a config tab people touch for all the wrong reasons just putting random numbers in.

mylar3:
merged up - thanks as always!

PR #1625, opened 2025-03-10
Added the ability to force search a subset of issues in the wanted list

falo2k:
Sometimes I don't always want to trigger a force search of my entire watchlist, but a subset that may cross over multiple volumes.  I could drill into each volume individually and initiate the searches, but that's too many clicks for my lazy fingers.  Instead I have repurposed the Force Check button to change behaviour when you have selections on the Wanted page, so it will just throw those items onto the search queue.

Behaviour:
- With no items selected, button behaves as before but is renamed to "Force Check All"
- With items selected, button is renamed to "Force Check Selected" and will only add the selected items to the queue

Useful for:
- Times I have added a load of backfill volumes for something new I want to read, and I want to initiate search across them all without also wasting search time on the items at the top of the watchlist that I know are just waiting for the standard release cycle.
- Initiating a search for a subset of Tier 2 items if I think they may have come available somewhere.

### DIFF
--- a/data/interfaces/default/upcoming.html
+++ b/data/interfaces/default/upcoming.html
@@ -6,7 +6,7 @@
 <%def name="headerIncludes()">
 	<div id="subhead_container">
 		<div id="subhead_menu">
-			<a href="#" id="menu_link_scan" onclick="doAjaxCall('forceSearch',$(this))" data-success="Checking for wanted issues successful" data-error="Error checking wanted issues">Force Check</a>
+			<a href="#" id="menu_link_scan" onclick="doAjaxCall('forceSearch',$(this))" data-success="Checking for wanted issues successful" data-error="Error checking wanted issues">Force Check All</a>
 		</div>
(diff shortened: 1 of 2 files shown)


evilhero (approved):

evilhero:
@falo2k think I did too many changes that were out of sync - got them back but this PR hits on previous changes due to webserve.py.

Can you do a new PR against the updated python3-dev branch and I'll merge it up no problems

Sorry, this is what I get for letting the PR's sit for awhile lol

falo2k:
No worries at all, it's not many lines to reproduce, plus I was thinking of changing the "All" to "Tier-1" on the button to be more correct.

I have noticed that the original Force Check doesn't add to the list if anything is currently in the search queue.  I wonder if that is confusing at all and should be tweaked, but I guess it's to avoid building up a search queue of endless duplicates, so probably a good thing.  Doubt many will notice / feel it.

PR #1631, opened 2025-03-14
Fixed the Mattermost notifier

Nebelherr23:
When I first poked around to write this, the notifier did not account for two things:

- imageFile being None (which would cause the notifier to barf)
- the base64-encoded string hitting a server-side max_body_size, resulting in a 400 - Bad Request and a log entry of "Premature EoF" on the side of the Mattermost instance

I decided to remove the image from the payload for these two reasons.

Alos this fixes the Test function (which didn't barf, but plain refused to work).

### DIFF
--- a/mylar/notifiers.py
+++ b/mylar/notifiers.py
@@ -484,7 +484,7 @@ class MATTERMOST:
         if module is None:
             module = ''
         module += '[NOTIFIER]'
-
+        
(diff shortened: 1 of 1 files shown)


Nebelherr23:
Oi, I hate git... Worked it against the wrong branch...

Issue #1564, opened 2024-08-25
Enforce File Extension Quality


falo2k:
**Describe the request**
Mylar does not currently enforce any ~~standards~~ common decency around file extensions - i.e. it doesn't ensure that RAR = cbr and ZIP = cbz.  A good example of this is when metatagging (i.e. implicit conversion to cbz): if it comes across a file that has a .cbr extension but is actually a zip container, it will leave the old extension intact.  It would be nice if it was a bit stricter and would sort that out.  Maybe, with time, we can naturally weed out the misnamed extensions, and nature can finally heal.

**How did you come up with the idea?**
I was just the right amount of annoyed with the latest encounter with a poorly named file, but I don't quite have the commitment to start up a squad of people to go around the world hunting down those who use bad extensions.  If I did, we would call them Ext-Force.

Nebelherr23:
You could make use of mylar's post-proc feature and just make sure it's the proper extension. (mimetypes are the keyword here)

Nebelherr23:
I forgot to mention that I thought of this as a work around.

falo2k:
Fair.  I did a mimetype pass through my entire library a while back to get everything aligned, and do the occasional manual scan for anything that's been left so it's doable as an exception process, but this came up tangentially in a support request on discord recently so figured would be good if it was one of those things we no longer needed to consider as a possibility when dealing with questions!

barbequesauce:
Obviously the boss can comment... from my perspective a number of us also use DC++ where we cannot modify the file released by the scanners. While we could modify a copy, space is not always a friend - leading to the preference for the (optional) post processing external script option rather than baked-in actions...

Cducharme84:
> Obviously the boss can comment... from my perspective a number of us also use DC++ where we cannot modify the file released by the scanners. While we could modify a copy, space is not always a friend - leading to the preference for the (optional) post processing external script option rather than baked-in actions...

You already presumably have the post-processing off otherwise Mylar would be changing the files, this suggestion is mostly geared towards correcting items that go awry during the post processing, in other words you would already be 100% unaffected as someone preserving hash.

Issue #1541, opened 2024-04-07
CRC Error Behaviour - Reject Comics


falo2k:
**Describe the request**
Can we get an option to reject comics when CT catches a CRC error?  It inevitably means a bad book and I have to either catch it in another application like Komga or when reading.  Would be much better if it just failed the issue so I could see it still sitting in my wanted queue as snatched and look into the problem.

**Environment (please complete the following information):**
 - OS: Docker/Unraid (LSIO)
 - Branch / commit: 9e2833974905d6ec18895fb640e2299eea5f7ec0

> 08-Apr-2024 00:11:48 - INFO    :: mylar.Process.757 : Thread-407 (Process) : Now checking: DCU: Legacies [33175]
> 08-Apr-2024 00:11:48 - INFO    :: mylar.duplicate_filecheck.2231 : Thread-407 (Process) : [DUPECHECK] Duplicate check for /media/downloads/usenet/comics/DC.Universe.-.Legacies.006.2010.Digital.Shadowcat-Empire/DC Universe - Legacies 006 (2010) (Digital) (Shadowcat-Empire).cbz
> 08-Apr-2024 00:11:48 - INFO    :: mylar.duplicate_filecheck.2389 : Thread-407 (Process) : [DUPECHECK] Duplication detection returned no hits. This is not a duplicate of anything that I have scanned in as of yet.
> 08-Apr-2024 00:11:48 - INFO    :: mylar.Process_next.2752 : Thread-407 (Process) : [POST-PROCESSING]  [1/1] Starting Post-Processing for DCU: Legacies issue: 6
> 08-Apr-2024 00:11:48 - INFO    :: mylar.run.147 : Thread-407 (Process) : ct_check: b'ComicTagger 1.3.5 [ninjas.walk.alone / SHURIKEN]\nCopyright (c) 2012-2024 ComicTagger Team\nDistributed under Apache License 2.0 (http://www.apache.org/licenses/LICENSE-2.0)\n'
> 08-Apr-2024 00:11:49 - WARNING :: mylar.run.285 : Thread-407 (Process) : [META-TAGGER][COMIC-TAGGER] file is not in a RAR format: DC Universe - Legacies 006 (2010) (Digital) (Shadowcat-Empire).cbz
> 08-Apr-2024 00:11:49 - INFO    :: mylar.run.223 : Thread-407 (Process) : [META-TAGGER] ComicRack tagging meta-tagging processing started.
[...]

Issue #1561, opened 2024-08-08
Automatic blacklisting of failed download


plumber-craic:
**Describe the request**
If an issue is snatched, downloaded, then deleted - that release should not be downloaded again. A hash of the downloaded file should be kept to enforce this. If a series is deleted the hash can be deleted too. 

**How did you come up with the idea?**
Ideally a feature to catch corrupt downloads would stop this happening in the first place, but I assumed this would be easier to implement. 

**Environment (please complete the following information):**
 - OS: Docker
 - Branch / commit: 0.8/master

**Additional information**
I also notice that the CBZ-only feature doesn't seem to work (the corrupt files I've seen have all been CBR)

plumber-craic:
possible duplicate of https://github.com/mylar3/mylar3/issues/1541

barbequesauce:
Chicken and egg - can't calculate the hash of the redownload until the (re)download is (re)done.

Issue #1543, opened 2024-04-20
Try post processing a 2nd time.


Dethkiller15:
**Describe the request**
When an initial post processing fails it attempts to extract and recompress the files and tries 1 more time.

**How did you come up with the idea?**
I have seen several times where you auto-download and manual download and it fails to process the downloaded file(s). Only way I have gotten it fixed was by manually extracting and making a new ".zip" file and rename ".zip" to ".cbz". Would be nice if that was just an option to have it done automatically. Also it has been happening a fair bit lately.

**Environment (please complete the following information):**
 - OS: docker
 - Branch / commit: latest

**Additional information**
Sometimes images are flat out corrupted in which case the program used to uncompressed should give an error for the corrupted files.

PR #1635, opened 2025-03-16
CRC checks on downloads and manual post-processing

falo2k:
Implemented file type checks and CRC checks for downloads from DDL, Usenet, Folder_Cache, and Manual Post-Processing.  Where possible the check is done straight after the download to enable "failed" marking, but for packs and manual that's not really possible so it happens in post processing.  That can cause a second pass of it, but it's harmless so wasn't worth branching for.

Currently, it will reject anything that isn't a PDF, zip, rar, or 7zip.  Can easily change that default behaviour if preferred.  It may help flag up failures with cloudflare as it'll read them as text files (no magic numbers) and reject them on quality grounds.

CRC checking is done for zip and rar using existing libraries.  7z and PDF are just passed if detected with valid header bytes.

Managed to find both bad rars and zips on usenet and GetComics to test against (Detective Comics (2016) is really really broken on usenet ...).

Resolves: https://github.com/mylar3/mylar3/issues/1541
Could be leveraged to resolve: https://github.com/mylar3/mylar3/issues/1564 but want to see how the newer CT handles those first.

### DIFF
--- a/mylar/PostProcessor.py
+++ b/mylar/PostProcessor.py
@@ -485,6 +485,24 @@ class PostProcessor(object):
                         if any(['coveronly' in cvchk, 'coversonly' in cvchk]):
                             logger.fdebug('Cover only detected. Ignoring result.')
                             continue
+
+                    if os.path.isfile(fl['comiclocation']):
+                        full_filename = fl['comiclocation']
+                    else:
+                        if fl['sub'] is not None:
+                            full_filename = os.path.join(fl['comiclocation'], fl['sub'], fl['comicfilename'])
+                        else:
+                            full_filename = os.path.join(fl['comiclocation'], fl['comicfilename'])
+
+                    # If after all that I still don't have a filename that is a file, something is probably confused.  Skip the integrity check.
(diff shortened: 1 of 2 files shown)


falo2k:
Hold on merging this for a bit.  Going to run in my daily driver and keep an eye on logs for anything unexpected.

falo2k:
As discussed, let's hold this until the CT merge, then I can rebase and squash.  Will run it in the meantime in case of any other path oddities.

falo2k:
Can maybe close the following with this as well as they result from corrupted file scenarios 
- https://github.com/mylar3/mylar3/issues/1561
- https://github.com/mylar3/mylar3/issues/1543

falo2k:
Been running this for a while on my main and it's been successfully ferreting out bad downloads, including those ones that tend to slip through the cracks with just a single slightly corrupt page in them.

falo2k:
Actually paid attention to PEP8 and squashed the commits.  If the force push has screwed with merging I can resub.

mylar3 (approved):

PR #1640, opened 2025-04-26
Fix: check NZBGet subpath is not None before adding it to the URL.

Sandluminat:
This PR fixes an issue where Mylar3 would incorrectly attempt to add the NZBGet subpath even when it was None, causing connection errors when the NZBGet Sub-Path was None.
In my docker installation it tried to access ip:6789 None instead of just ip:6789

### DIFF
--- a/mylar/nzbget.py
+++ b/mylar/nzbget.py
@@ -41,7 +41,7 @@ class NZBGet(object):
             return {'status': False}
         url = '%s://%s:%s'
         val = (protocol,nzbget_host,mylar.CONFIG.NZBGET_PORT)
-        if mylar.CONFIG.NZBGET_SUB:
+        if mylar.CONFIG.NZBGET_SUB is not None:
             if not mylar.CONFIG.NZBGET_SUB.startswith('/'):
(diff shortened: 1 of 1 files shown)


falo2k:
Thanks for the explanation of the issue, but is the empty value subpath not already solved by [this commit](https://github.com/mylar3/mylar3/commit/b82dc317293dbd071c00365669c34c895317b302) in the dev branch?  I can't reproduce the malformed URL with a blank entry when running against that branch.

> It's good instinct, but I think its implementation is incomplete.
I think "incomplete" is an understatement.  The changes in this PR do not change the behaviour for a value of None, and if for some reason a blank value reached that point, would no longer skip that either.  A string value of 'None' (again, should it reach this point) would make no difference before or after this change.  You could argue the toss over style / risks of other truthy values, but I don't believe that's the intent of this change, and it's hardly the only non-PEP8 friendly condition.

> This should probably account for users who might have their configuration accessible via redirect to a host address, without a port. Example: http://nzbget.lan/

There is still a port there, it's just the default for that protocol.  I'd hope that if someone is setting up such a redirect, they'd understand that already, but then there's always a better moustrap.  Could default it based on the protocol for the host.

PR #1644, opened 2025-06-09
AirDC++ for Search & Download

plumber-craic:
Initial implementation of using AirDC++ Web to search and download. Also monitors release announcement hubs as if they were RSS feeds. Likely there are some things I am not doing correctly as I am not that familiar with the Mylar codebase, but I have tested it and verified it can search, download and post process, and that rss force + schedule work. I put it in Search Providers near DDL - it is both search and download so wasn't sure. 

My test environment was a dockerised verlihub and a pair of airdc++ containers. I used one container to be the search endpoint and the other to share the comics. Both connected to the test hub. I then created an API user in AirDC++ under Settings, System, Add User and put those credentials into Mylar. 

Features:
* Hub filter - important in case you don't want all your hubs to be searched by Mylar
* Search - creates a search instance, searches the hub with this instance, waits for it to get results
* RSS for Announce Hubs - use a release announcement hub like an RSS feed
* Download - this gets the TTH of the result and queues the bundle for download
* Post Process - leverages existing mylar post processing. Tested with Copy and Move
* Test button with version info

image

### DIFF
--- a/data/interfaces/default/config.html
+++ b/data/interfaces/default/config.html
@@ -834,6 +834,58 @@
                         </fieldset>
                         <fieldset>
                                 <div class="row checkbox left clearfix">
+                                     <input id="enable_airdcpp" type="checkbox" onclick="initConfigCheckbox($(this));" name="enable_airdcpp" value="1" ${config['enable_airdcpp']} /><legend>AirDC++</legend>
+                                </div>
+                                <div class="config">
+                                   <div id="airdcpp_options">
+                                       <div class="row">
+                                          <label>Address:Port</label>
+                                          <input type="text" name="airdcpp_host" id="airdcpp_host" value="${config['airdcpp_host']}" size="30">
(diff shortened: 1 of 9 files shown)


faethon:
Any chance this gets merged into main? This looks to be a very useful addition.

mylar3:
Since this was forked originally off the master branch, and it's a huge PR with the same file(s) being touched - can you do a clean PR against the dev branch? When you fork off the master and then switch branches after the PR it messes up the merge for us later on when it goes back into master and we tend to get merge errors related to the same files being touched repeatidly and it becomes a very long process to get past that.

Outside of the fact that I don't personally use dc++, support is gonna be awkward for this - but I know that lots of users have asked for this. So once you can get a clean PR up, I'll merge it up ASAP.

And if it wasn't mentioned previously, much appreciation for tackling this!

(Going to close this just to make room for the future clean PR)

PR #1647, opened 2025-06-21
Update requirements.txt to add pyOpenSSL for SSL

Salvora:
Currently, SSL is not working due to missing dependency pyOpenSSL in the requirements.


barbequesauce (changes requested):
Apologies, we do not accept PRs against master - only dev. We appreciate the offering however! This change would be likely successfully reviewed for dev and we encourage you to re-submit!

PR #1654, opened 2025-06-28
Fix exception on single wanted issue submit, revert button text on refresh

falo2k:
I had missed a very obvious case of selecting a single issue on the wanted page for checking causing the issueIds to not be passed as a list.  Clearly I've been using the magnifying glass too much ...


mylar3:
I think we'll need a rebase on this against the current branch before we can merge up.

PR #1659, opened 2025-07-10
Improved issue number handling

falo2k:
This is the previous change we talked about for a single consolidated function for issue number processing, with the other side of the equation finally closed.  Either filechecker isn't as bad as you warned, or I have gone insane as I think I'm starting to get it.

Rather than trying to eliminate the issue exceptions list, I have instead embraced it and automated it.  The list is now made up of a combination of a hard-coded list and one stored in config.  The config one can be added to manually, but it will also be updated automatically whenever the issue number processing function is called on a CV data update and spots something that isn't covered yet.  Future updates can move candidates into the hard coded list, and the config will clean itself up (within reason - I can see an improvement where it differentiates between patterns and absolute strings, but don't think it's an urgent change).  Demonstrated this with ASM and issue 65.Deaths - it could process ASM and adopt the Deaths exception without being told it needed to.

I've added in some additional changes for supporting specific issue number patterns (like Cerberus with 289/290 or 112-113), and other special cases like non-numeric One-Shot issue numbers.

This should now cover a bunch of volumes that previously either would never load properly, could never match files properly, or would rename themselves into an archived state.  

I've tested against a list of oddity volumes on a test instance, and it seems solid, but given that particular number parsing function touches **a lot** of places it at least warrants some extra eyes on the changes.

Have kept a small change to logging as a seperate commit as it's distinct from the purpose of the PR, but was essential to banish some exception noise.


falo2k:
Slept on this and decided that making the distinction between pattern and exact matches was going to lead to a cleaner approach overall.  Appreciate will likely need to rebase this and resubmit given changes over the horizon so can squash the changes then - might also PEP8 the changes when I do.

mylar3:
this one needs a rebase against the updated dev branch as the helpers file was already modified earlier and it's throwing conflict errors.

PR #1660, opened 2025-07-11
FIX: More stuck SAB downloads due to PP statuses

falo2k:
Some people (myself included) have reported downloads still getting stuck.  Managed to catch one that fell through the net - it was in the Verifying state.  I've added some more status codes to your previous fix to account for this.

Selected status codes based on the PP status strings here: https://github.com/sabnzbd/sabnzbd/blob/872cf835dfa1daeec43cddb0bc8377434c14bfde/sabnzbd/constants.py#L168

I left Fetching off the list as from reading my assumption is that should end up back in the queue API call.  I may be wrong.  Going to run this as my daily for a week or so to see if any still get stuck.

### DIFF
--- a/mylar/sabnzbd.py
+++ b/mylar/sabnzbd.py
@@ -276,7 +276,7 @@ class SABnzbd(object):
                 elif hq['nzo_id'] == sendresponse:
                     nzo_exists = True
                     logger.fdebug('nzo_id: %s found while processing queue in an unhandled status: %s' % (hq['nzo_id'], hq['status']))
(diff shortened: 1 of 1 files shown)


falo2k:
FYI, been running this in combo with #1665 #1663 #1662 #1660 #1635 and #1636 for the last week.  Been solid, and the only missed post porcesing I've had have either been explicit fails from CRC checks, or DDL mismatches/gaps.

mylar3 (approved):

PR #1662, opened 2025-07-14
FIX: SQLite3 single quoted string literals

### LINKED ISSUE #1641: Double quoted string literals in sqlite
Noticed through a user having issues with database queries in freebsd.  Looks like at some point sqlite tightened up on flexibility around double or single quoting string literals.  Actual strictness then varies by package - a browse around suggests Fedora went strict at some point, FreeBSD did, Ubunutu seems still flexible at the moment.

Links discussing issue:
https://www.reddit.com/r/freebsd/comments/1chb82b/comment/l246a68/
https://www.sqlite.org/quirks.html#dblquote

Could either go through and clean up all queries, or potentially set a config option on the database to allow double quoting (https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.setconfig with sqlite3.SQLITE_DBCONFIG_DQS_DML).

falo2k:
This is an attempt to resolve #1641 without resorting to changing the sqlite config in a future python version.

- Hunted down as many uses of double quoting for string literals in queries as I could with some regexfu in an attempt to restore compatibility for FreeBSD / prepare for any future risk of sqlite3 quote support changing in images.  In general just took the approach of switching query strings from single quoted python strings to double quoted python strings (to avoid escapes for the sake of escapes).  In some cases did some cleanup for readability (concatenation -> f-strings).
- Fix in search.py (searchforissue) to a couple of queries that were malformed.  Based their fix on the logic used in a previous loop case.

### DIFF
--- a/mylar/api.py
+++ b/mylar/api.py
@@ -342,20 +342,20 @@ class Api(object):
         else:
             self.id = kwargs['id']
 
-        comicQuery = '{select} WHERE ComicID="{id}" ORDER BY ComicSortName COLLATE NOCASE'.format(
+        comicQuery = "{select} WHERE ComicID='{id}' ORDER BY ComicSortName COLLATE NOCASE".format(
             select = self._selectForComics(),
             id = self.id
         )
         comic = self._resultsFromQuery(comicQuery)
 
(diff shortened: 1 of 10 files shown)


falo2k:
Re-e-base. When the crowd say bo selecta.

falo2k:
Superceded by #1703

Issue #1664, opened 2025-07-15
Occasional issues getting confused with CDH


falo2k:
**Describe the bug**
I've had a couple of these failed post-processing SAB downloads over the last couple of days.  I cannot reproduce them, and they don't appear to be linked to the specific NZB as it will redownload and process fine.  There's no other issue with the download, and hundreds of others are unaffected.  What's particularly odd is that the output from the logs would suggest that the value returned in the history queue by SAB for <storage> is "." (only reason I can see why it would get to that point), but if I check the history API myself it is what I'd expect (a proper, and absolute, path).  It can't (shouldn't?) be another problem with SAB states as it must have been Completed to hit this patch of code, and it confirms as much in the logs.  There's nothing that stands out in the SAB logs for that download.  I'm properly stumped.

The only last thing I can think is that the SAB status change isn't atomic with other updates to the history queue and for some reason the path ends up temporarily incorrect in a fraction of a ms, but I wouldn't expect to see that with a frequency of ~1 in 150 snatches.

** Mylar Log Snippet of CDH for that NZO **
```15-Jul-2025 21:16:31 - DEBUG   :: mylar.processor.96 : AUTO-COMPLETE-NZB : successfully queried the queue for status
15-Jul-2025 21:16:31 - WARNING :: mylar.processor.102 : AUTO-COMPLETE-NZB : Unable to locate item within sabnzbd active queue - it could be finished already?
15-Jul-2025 21:16:31 - DEBUG   :: mylar.processor.105 : AUTO-COMPLETE-NZB : SABnzbd Queued item status : Idle
15-Jul-2025 21:16:31 - DEBUG   :: mylar.processor.106 : AUTO-COMPLETE-NZB : SABnzbd Queued item mbleft : 0.00
15-Jul-2025 21:16:31 - INFO    :: mylar.processor.145 : AUTO-COMPLETE-NZB : File has now downloaded!
15-Jul-2025 21:16:31 - DEBUG   :: mylar.historycheck.170 : AUTO-COMPLETE-NZB : SABnzbd version is higher than 3.2.0. Querying history based on nzo_id directly.
15-Jul-2025 21:16:32 - DEBUG   :: mylar.historycheck.188 : AUTO-COMPLETE-NZB : nzo_id: SABnzbd_nzo_re_g4r0t --- SABnzbd_nzo_re_g4r0t [Completed]
[...]

PR #1665, opened 2025-07-16
FIX: Check for valid storage response from SAB

falo2k:
Suggested quick fix for #1664.

Could put a more robust failure state to drop out if it happens on roundtwo, but it will go on to fail in later passes anyway and this is to account for what I believe to just be a blip in SAB responses.


falo2k:
A fix has gone in for this to the SAB develop branch, but it wouldn't hurt to retain the check for people will older installs / in case there's another timing issue hiding out there somewhere.

mylar3 (approved):

PR #1673, opened 2025-07-24
Expose HOST_RETURN value for SAB configuration in UI

falo2k:
I figure with docker deployments becoming more commonplace this might be a good idea to expose (and how many times have people repeated the phrase "You'll need to change host_return in config.ini.  Make sure that mylar is shut down when you do so"?

Moved the routine for calculating the return address for mylar into a standalone helpers function so that I could use it to render the "detected" address on the UI when exposing the HOST_RETURN config value (acts as both a style hint, and information).

Doesn't do any validation on the entered value.


falo2k:
Superceded by #1703

PR #1675, opened 2025-07-26
Added status_code to log entries for getcomics searches

falo2k:
Thought it might be handy to a) have the status code at a glance and b) give us a chance to check whether we're getting a [custom status code](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/) back from CF in the case of likely blocking.  Could then potentially add some handling / notification later for the end user.


falo2k:
Tested CloudFlare blocking using a free ProtonVPN account.  Search page gets 403ed and, after inspecting the request contents, it is indeed a CloudFlare block.  Since it's possible there could be failures for other reasons (site down, broken links, temporary connectivity, user has left their VPN on, etc.) I didn't think it warranted anything more aggressive than putting a more informative log message.  Should at least make those incidents easier to diagnose in support chat.

PR #1678, opened 2025-07-29
Save the currently selected letter and paging for alphaindex mode

### LINKED ISSUE #1613: Go back to original position when navigating back from individual series screen
**Describe the request**
Having a bit of a hard time making this clear: When I select a letter (e.g. "G") to see all the series with that letter, and I then select a series from the list, I'd like to be able to hit the back button on the browser and return to the same selection (in this case "G") that I left. Instead it returns to the full list, although sometimes it'll choose a random letter that was used prior. I haven't been able to establish a pattern for that yet.

**How did you come up with the idea?**
It's simple annoyance with having to choose the same letter over and over then navigate to the same page within that group over and over.

 - OS: Docker on Ubuntu 24.10 VM
 - Branch / commit: Dev

**Additional information**
I see that it returns to the monolithic list of all series regardless, but making some kind of internal bookmark to return to would be amazing.

falo2k:
- Stores the current alphabet index selection in local storage and restores on page load
 - Orange text colour for selected Alpha page if 0 items to differentiate between other empty pages
 - Default to page 0 (None) on first load / with no state

Resolves #1613

### DIFF
--- a/data/css/alphabetSearch.css
+++ b/data/css/alphabetSearch.css
@@ -25,6 +25,10 @@ div.alphabet span.empty {
         color: red;
 }
 
+div.alphabet span.empty.active {
+        color: orangered;
+}
+
 div.alphabetInfo {
         display: block;
         position: absolute;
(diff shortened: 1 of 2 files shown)


falo2k:
Closing this in favour of #1686

PR #1681, opened 2025-08-04
FIX: Story Arcs with annual issues didn't work when annual integration is on

EricShell:
When annual integration is enabled, a story arc issue is an annual, it will sometimes match issues and sometimes will not. I debugged this, and found that this one line was not returning an issue in these cases. Looking at the DB, it was failing to match up the id between the annuals and storyarcs tables correctly. It was looking at the annuals table's comicid (which is the main comic series id) instead of the releasecomicid (which is the annual series id). Changing this to releasecomicid causes it to properly match.

After this change, I stood up the server locally with my entire DB, and tried searching for missing on several story arcs, some with annuals and some without. I also tried refreshing several story arcs. Saw no errors or issues, and the bug was fixed.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -1986,12 +1986,12 @@ def listLibrary(comicid=None):
     myDB = db.DBConnection()
     if comicid is None:
         if mylar.CONFIG.ANNUALS_ON is True:
-            list = myDB.select("SELECT a.comicid, b.releasecomicid, a.status FROM Comics AS a LEFT JOIN annuals AS b on a.comicid=b.comicid group by a.comicid")
+            list = myDB.select("SELECT a.comicid, b.releasecomicid, a.status FROM Comics AS a LEFT JOIN annuals AS b on a.comicid=b.comicid group by a.comicid, b.releasecomicid")
         else:
             list = myDB.select("SELECT comicid, status FROM Comics group by comicid")
     else:
         if mylar.CONFIG.ANNUALS_ON is True:
-            list = myDB.select("SELECT a.comicid, b.releasecomicid, a.status FROM Comics AS a LEFT JOIN annuals AS b on a.comicid=b.comicid WHERE a.comicid=? group by a.comicid", [re.sub('4050-', '', comicid).strip()])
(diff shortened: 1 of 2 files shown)


EricShell:
Found a couple of issues when doing further testing:
1. The link from the storyarc page to the comic in your library isn't working for some annuals. Need to investigate why, but I suspect it's another place where the join isn't fully working.
2. I found one case where it is still not matching the comic from my library for a story arc. I suspect it's just a data issue, but I want to track that down and see if I can fix that cleanly as well.

Will move this back off of draft and put another comment once I have it sorted.

falo2k:
Just checking for the second one, is it possible it's pointing to a different but very similar volume (foreign release, TPB, special edition, etc.)?  The ComicVine StoryArc data is a bit rubbish.

EricShell:
Yeah I suspect it is down to non-consistent data as well, and in my particular case it's an annual where I have the annual volume added directly, not via annual integration. I am hoping I can make the join robust enough to handle this, but will see once I get into the actual data and debugging.

EricShell:
I fixed #1 from the above comments, and it also fixed a similar bug affecting search results. The issue was when a single series had multiple annuals, and that is now working well. Still working on #2.

Also it looks like my commits got all messed up when I rebased my local branch to include all the merged changes from dev. Will work on cleaning that up too.

caddy-wampus:
This is ready to go now - I got the commits cleaned up nicely.

For the 2nd issue I mentioned above, it only comes up for one particular annual that I have added manually, and that annual has some strange naming. I found where things are going wrong in the code, but this is going to be a larger change, so I will do that in a separate PR (this problem was not introduced nor worsened in this PR, and will only come up rarely).

caddy-wampus:
Filed issue #1704 to track the third issue above (and to facilitate any discussion needed on that issue). I will work on a fix for that next.

evilhero (approved):

evilhero:
Everything looks good from what I can see - not sure why we were matching on the incorrect id when annuals was on (probably an errant cut & pates I'm guessing). As far as the matching to the name, there was some reasoning behind it, but atm I can't think of one and if it's working as intended with this PR I'm ok with forgetting the reason why.

Thanks for the PR and the detailed explanation as to the why.
Cheers!

PR #1684, opened 2025-08-12
Support multi-column sorting on non-alphaindex home page

falo2k:
Multi column sorting for the volume datatable on the home screen (to enable sort by Comic, then Year)


falo2k:
Closing this in favour of #1686

PR #1685, opened 2025-08-13
FIX: Incorrect totals for All filter on search results page

falo2k:
The -1 page size for "All" in DataTables was causing list truncation and dropping the final record.


falo2k:
Superceded by #1703

PR #1690, opened 2025-08-16
Ability to add an instance name to the browser title

falo2k:
One of those "didn't know I wanted it until I wanted it features".  Possibly useful for those who run multiple instances of mylar for manga/western/publisher distinctions.  Definitely useful for me when I run multiple for different versions, branches, config, etc.

Just adds a configurable string between "Mylar" and the page name in the title bar (e.g. `Mylar - Dev Instance - Home`).


falo2k:
Superceded by #1703

PR #1696, opened 2025-08-17
defaults in config.py had hardcoded strings instead of empty strings

plumber-craic:
defaults in config.py had hardcoded strings instead of empty strings


mylar3:
thanks for the PR - but was reworking the config for the dc+ stuff and this was already in there. Instead of having you rebase et all again, I just left my changes and this is now in python3-dev.

Cheers

PR #1702, opened 2025-08-19
IMP: moved airdcpp variables into own section of config, IMP: removed '#' from airdcpp search terms

evilhero:
config will update to v15 and auto-convert existing airdcpp values into the new sections and reuse them immediately.

removed ``#`` from the airdcpp query search terms as including it garnered zero hits until the last iteration of searching (3 loops usually not results, fourth results).


mylar3 (approved):

mylar3:
supercedes #1696

PR #1715, opened 2025-08-23
Update api.py to set issue Archived status

ACDChook:
Use the named but unused 'addIssue' endpoint to set status of an issue to Archived.

### DIFF
--- a/mylar/api.py
+++ b/mylar/api.py
@@ -796,6 +796,18 @@ class Api(object):
         newValueDict = {'Status': 'Skipped'}
         myDB.upsert("issues", newValueDict, controlValueDict)
 
+def _addIssue(self, **kwargs):
+        if 'id' not in kwargs:
+            self.data = self._failureResponse('Missing parameter: id')
+            return
+        else:
+            self.id = kwargs['id']
+
+        myDB = db.DBConnection()
+        controlValueDict = {'IssueID': self.id}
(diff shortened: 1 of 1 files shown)


falo2k:
Hi.  PRs should be against python3-dev, and the indentation is incorrect in your function definition.

Also, why?

ACDChook:
I can recreate it against python3-dev.

I just copied the formatting directly from the existing api.py code.

It seems incomplete to not be able to set the status of an issue using the API to any of the statuses that can be set with the web UI. And the 'addIssue' endpoint was named in the file already, but had no code for it.

evilhero:
Probably because it's for adding an issue which isn't possible as it adds by series, not issue. 

If you were going to do a PR for archiving an issue, you would create a relevant endpoint name not some name already declared  - used or not.  As an aside, there's also no checks to verify that it's just an issueid / status being sent and not something entirely different. 

All that being said, there's already ``changeStatus`` so I don't see the need or use for this.

falo2k:
You may have copied it, but you clearly pasted it incorrectly.
image

If you want to be able to set the states, it would make sense to implement an additional API endpoint to handle that with sensible naming.  Having "addIssue" change the state to Archived is unintuitive.  I'd also want to see some guards against the issue actually being there before just calling an upsert.

== Edit: Beat me to it.

PR #1716, opened 2025-08-23
Add "archiveIssue" API endpoint to api.py

ACDChook:
Definitions exist for setting individual issue status to Wanted or Skipped. Add an endpoint for setting an issue to Archived as well.

changeStatus only changes all issues in a comic series at once, which is not desired behaviour if only one single issue needs updating.

Formatting has been copied directly from other definitions for consistency.

### DIFF
--- a/mylar/api.py
+++ b/mylar/api.py
@@ -34,7 +34,7 @@ import datetime
 
 cmd_list = ['getIndex', 'getComic', 'getUpcoming', 'getWanted', 'getHistory',
             'getLogs', 'getAPI', 'clearLogs','findComic', 'addComic', 'delComic',
-            'pauseComic', 'resumeComic', 'refreshComic', 'addIssue', 'recheckFiles',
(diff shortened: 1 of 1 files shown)


evilhero:
You keep on mentioning that you need to set the status to archived, but what are you doing that it's automatically not being set? 

If it's in a ``Downloaded`` status, deleting the file would put it into an ``Archived`` status. 

Marking it as ``Archived`` if the file exists in the series directory won't change the status because it will get flipped to ``Downloaded`` on the next Refresh/Recheck.

If you just want to know the file isn't available, there is the ``Ignored`` status as well.

Not a fan of this layout, as it should be inclusive for all of the status that doesn't fire off a job (ie. Wanted), allow for batch id's and have some sanity checks for data (basically the changeStatus with some improvements)

ACDChook:
If an issue is found manually and isn't downloaded by mylar, setting the status of the issue to Archived so it's consistent with other issues found by mylar seems logical. And I've got comic files in a separate file structure that mylar doesn't access directly, so issues that I have stay at Archived status.

barbequesauce (approved):

PR #1717, opened 2025-08-23
FIX: Fix to corner case of numbering performance tweaks

falo2k:
Restores some of the previous checks that it turns out were in place for a reason, but keeps them after the isdigit early exit to maintain performance.


falo2k:
The additional commit reverts some behaviour with filechecker to how it was before (while still retaining the dynamic exception list).  This does unfortunately remove (or revert) the support for parsing string prefix issues (Weapon Zero T-2) and pure string issues (Earth X X).  Back to the drawing board for those ,but probably after writing some unit tests.

evilhero (approved):

PR #1722, opened 2025-09-02
Tracking branch august25

falo2k:
I wasn't going to push this now, but if you wanted something innocuous to push to sort out the refs->carepackage problem ...
- Cleans up the rogue .orig file I committed in a patch and adds .orig to gitignore (it's a remnant of git's mergetool)
- On the subject of cleanups, we can delete comicdetails.html right?  Would certainly help to clean up codebase searches
- Added a quick string clean to carepackage generation where there are multiple refs in case it happens again.  Concatenate them, and clean up some standard bad-filename characters.


falo2k:
Running this in my main env and it just survived the first hit of weekly RSS updates without leaving me with a load of extra downloads.

falo2k:
Fix is for case where folder monitor still triggers because of a previously populated CHECK_FOLDER value (and it was the /comics folder).

falo2k:
Will resubmit as a new branch after rebase

PR #1727, opened 2025-10-09
[Feature] Add configurable comicvine URL

barrelltitor:
Useful if you want to use a cache like mine. Pretty simple addition. Tested it on a dev instance

### DIFF
--- a/lib/comictaggerlib/comicvinetalker.py
+++ b/lib/comictaggerlib/comicvinetalker.py
@@ -77,6 +77,7 @@ class ComicVineTalker(QObject):
 
     logo_url = "http://static.comicvine.com/bundles/comicvinesite/images/logo.png"
     api_key = ""
+    api_base_url = "https://comicvine.gamespot.com/api"
 
     @staticmethod
     def getRateLimitMessage():
@@ -88,7 +89,10 @@ class ComicVineTalker(QObject):
     def __init__(self):
         QObject.__init__(self)
 
-        self.api_base_url = "https://comicvine.gamespot.com/api"
+        if ComicVineTalker.api_base_url in ("", None):
+            self.api_base_url = "https://comicvine.gamespot.com/api"
+        else:
+            self.api_base_url = ComicVineTalker.api_base_url.strip().rstrip('/')
         self.wait_for_rate_limit = False
 
         # key that is registered to comictagger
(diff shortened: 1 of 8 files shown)


lib/comictaggerlib/comicvinetalker.py
```diff
@@ -88,7 +89,10 @@ def getRateLimitMessage():
     def __init__(self):
         QObject.__init__(self)
 
-        self.api_base_url = "https://comicvine.gamespot.com/api"
+        if any(ComicVineTalker.api_base_url == "" or ComicVineTalker.api_base_url == ""):
```
barrelltitor: Not sure what happened here, I think I meant to put is None, will fix

lib/comictaggerlib/comicvinetalker.py
```diff
@@ -88,7 +89,10 @@ def getRateLimitMessage():
...
+            self.api_base_url = "https://comicvine.gamespot.com/api"
+        else:
+            self.api_base_url = ComicVineTalker.api_base_url.strip().rstrip('/')
```
barrelltitor, line 95: here I don't know if stripping is needed, figured since this one was without a final slash might as well strip it just in case

mylar/config.py
```diff
@@ -1360,6 +1361,17 @@ def configure(self, update=False, startup=False):
             # Set the actual API key, so mylar does not appear broken from the start
             self.COMICVINE_API = self.COMICVINE_API[4:]
 
+        # Ensure ComicVine URL is always set and normalized according to previous url 
+        if any([self.COMICVINE_URL is None, self.COMICVINE_URL == '', self.COMICVINE_URL == 'None']):
+            self.COMICVINE_URL = 'https://comicvine.gamespot.com/api/'
+        else:
+            self.COMICVINE_URL = helpers.clean_url(self.COMICVINE_URL)
+            if not self.COMICVINE_URL.endswith('/'):
+                self.COMICVINE_URL += '/'
```
barrelltitor, line 1370: not sure if this part is needed, but I can remove it if not. shouldn't really harm anything though

evilhero:
I think I'd much prefer the option to not be  available in the GUI, just because the GUI is already confusing enough as is without adding a power-user option that realistically 99% of the userbase wouldn't use (atm at least).

Outside of that, the PR is fine. If it can be removed and the commits squashed into one, I have zero problems merging it in.

Just quickly glanced - the passing of the cv url in each call seems abit unnecessary since if it's not the default there's no reason to include it in each call.

barrelltitor:
Makes sense, will get all of this done tomorrow

barrelltitor:
Done, dropped the UI part and made it only add the tag option in case it's not the default URL

barbequesauce (approved):

PR #1728, opened 2025-10-11
[Feature] Send DDL downloads to JDownloader2

barrelltitor:
So this might need the options removed from the UI as well since this would be more of a power user feature, but this allows you to send off downloads to jdownloader 2  and monitor their completion

It picks the queue back up on startup and automatically fails any failed or unstarted downloads

It sends all mirrors to jdownloader so it can pick the best one, this solves the issue where the best link is dead

I saw there's an external downloader option somewhere in the code but I found nowhere in the code actually using it, it seems to be just a random variable

I finally started using mylar for ddls after adding this in, works like a charm so far

### DIFF
--- a/data/interfaces/default/config.html
+++ b/data/interfaces/default/config.html
@@ -912,6 +912,23 @@
                                                 <input type="checkbox" id="ddl_prefer_upscaled" name="ddl_prefer_upscaled" value="1" ${config['ddl_prefer_upscaled']} /><label>Prefer HD-Upscale/HD-Digital <small>(if available></small></label>
                                            </div>
                                        </div>
+                                       <div class="row checkbox left clearfix">
+                                            <input type="checkbox" id="jd2_enable" onclick="initConfigCheckbox($(this));" name="jd2_enable" value="1" ${config['jd2_enable']} /><label>Send to JDownloader2</label>
+                                       </div>
+                                       <div class="config">
+                                           <div id="jd2_options" style="display:none">
+                                               <div class="row">
+                                                   <label>JDownloader2 URL</label>
+                                                   <input type="text" name="jd2_url" id="jd2_url" value="${config['jd2_url']}" size="30">
+                                                   <small>(ex. http://127.0.0.1:3128)</small>
+                                               </div>
+                                                <div class="row">
+                                                    <label>Download Path</label>
+                                                    <input type="text" name="jd2_dest_dir" id="jd2_dest_dir" value="${config['jd2_dest_dir']}" size="30">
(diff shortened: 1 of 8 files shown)


barrelltitor:
I'd love some more testing on this and opinions on design decisions or etc. I think this is very useful to have since jdownloader2 is quite a popular download manager

This might need to come with some instructions somewhere in the docs on how to properly setup the jdownloader side as well, as it needs turning on the local API and exposing its port

barrelltitor:
JD2 Instructions:

- Settings -> Advanced -> Search for "Remote API"
- Enable Deprecated Api
- If using docker either set mylar and jd2 in the same network, or (unsafe) Disable "Deprecate api localhost only"
- Choose a port and make sure that port is exposed in your jd2 instance however you host it

The "deprecated" API is just the local-version of their new "myjd" api, which is cloud only. It has no protection, so make sure not to expose this under any circumstances to the internet. 

In Mylar:

- Enable JD2
- Set the JD2 URL to http://jd2_IP:PORT (hostname can be used as well , just make sure to add the port or reverse proxy it)
- Set the JD2 download destination. This location must match in both mylar and JD2, so it must be mapped the same if used in containers

barrelltitor:
Squashed everything. It works fine for me, I've been using it ever since, but I would like some more testing if anyone could do it. 

Otherwise it should be good at least from my end, will wait for the review from evilhero

barrelltitor:
Okay so this changed a bit. The way it currently works:

- getcomics.py does its normal thing and finds the links
- instead of sending the download to the ddl queue we
  - make another list of links from tmp_links and assign a priority based on the config
  - send the list to jdownloader 2 and add it to the jd2_queue
    - if there is an error, fallback to ddl queue
  - jdownloader2.py sends each link in quick succession, assuring the download is not started until all links are received and assigns a job_id
  - the jd2 queue handler checks each job ID for a completion state
    - if the state is completed, it sets that in the ddl_info and sends it to the pp_queue if enabled
    - if the state is not yet completed, add the job back to the queue and sleep
    - if the state is failed, fail it 
    - if the download never started within 3 minutes(should probably make that a configurable option)   assume it's in a failed state and fail it

Edge case: if you send many comics to download in quick succession this will block anything not yet moved to the download queue from being moved there. The solution would probably be to move the jd2 submit function inside the queue and sleep there to allow time for it to start, then add it back in the queue for monitoring, though I got this to happen by redownloading the same files so the html cached kickeded in

barrelltitor:
It's pretty much done, just waiting for further testing and reviews

barbequesauce (approved):

PR #1733, opened 2025-10-14
FIX: DDL RSS fix to skip non-file posts

falo2k:
The DDL feed processing was throwing an exception because of the missing header paragraph in non-file posts.  Not sure if other non-file posts have shown up in the feed before or if it normally filters (the culprit in this case was tagged as "News" whereas previous things like site updates are tagged as "Blog").  Regardless, will skip an entry if it doesn't have that expected centered paragraph of year + size data.


falo2k:
Closing as merged under #1745

PR #1734, opened 2025-10-15
DEV: Development tools (GH actions, devcontainers, tests)

falo2k:
Right, offering this up for consideration.  A collection of things I've been playing around with in the background:
- Github actions to automatically test any push to a non-core branch (i.e. not master or python3-dev) against python 3.8-3.11 on both Windows and Linux.  Checks out the repo and launches Mylar, checking that it can then access **something** at localhost:8090.  Could be enhanced with deeper tests in the future, but initially works as just a smoke test to prove deployment works.
- A github action (same conditions as above) that will build and publish to GHCR a Docker image based on the LSIO nightly Dockerfile.  As well as being based on the pushed branch, the container also runs debugpy and opens a port so that you can connect remotely to debug the container.  I can write up steps to set up a simple launch.json profile for VS Code to do that connection.
- A [devcontainer ](https://devcontainers.github.io/) configuration that enables building of a local docker container to run mylar.  VS Code makes it really easy to launch and build (even with the pain of Docker Desktop), so handy for quickly testing a linux build on a Windows dev machine.
- A start on implementing some pytest coverage (and a triggerable GH action to run them - currently on a manual trigger).  So far I've covered a large chunk of helpers.py with unit tests, and a test on filechecker.py for parsing as an example.  The not-yet covered functions in helpers.py are mostly those where they'll need some database mocking and I'm still thinking about how I'd like to approach those.  I'd quite like to put together a fixed dataset (just structured in json) that could be used to both drive a mock/test sqlite database and provide mock CV API responses for testing.  I haven't decided whether I want to re-use that for unit tests stubs vs some integration test mocks (down the road).

I'm not suggesting mandating test coverage, but will aim to slowly try to build up the test suite across the codebase in the background.  I will also write up a new page for the mylar docs site that covers these things.

### DIFF
--- a/.devcontainer/comics/README.md
+++ b/.devcontainer/comics/README.md
@@ -1,0 +1,1 @@
+Default comic storage location mapped to the devcontainer
--- a/.devcontainer/config/README.md
+++ b/.devcontainer/config/README.md
@@ -1,0 +1,1 @@
+Default mylar config path mapped to the devcontainer
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,35 @@
+{
+    "image" : "mcr.microsoft.com/devcontainers/python:3.8",
+    "customizations": {
+        "vscode": {
+            "extensions": ["ms-python.flake8", "ms-python.debugpy"],
+            "settings" : {
+                "launch" :{
+                    "version": "0.2.0",
+                    "configurations": [ {
+                            "name": "Debug Mylar in Dev Container",
+                            "type": "debugpy",
+                            "request": "launch",
+                            "program": "Mylar.py",
+                            "console": "integratedTerminal",
+                            "args": ["--datadir", "/config/"]
+                        }
+                    ]
+                },
+                "remote.autoForwardPorts": false
+            }
+        }
+    },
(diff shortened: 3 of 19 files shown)


falo2k:
Forgot to mention - comments at the bottom of the test_helpers.py file capture:
1. Functions not yet covered by testing (for the reasons above)
2. Functions not covered that I think could be reviewed for refactoring or deletion because they're either unused, superceded, or could do with some quick tweaks.

aptalca:
Just a quick heads up, the lsio docker image is based on Alpine 3.19, due to python 3.11 being the latest supported python version by mylar3. Unfortunately, Alpine 3.19 will be EOL in a couple of weeks (Nov 1, 2025).

Most of our alpine based images are currently on Alpine 3.22. Mylar3 image was held back due to python. We usually deprecate our baseimages when the distro version goes EOL. Not sure what to do with Mylar3 when that happens. Maybe rebase to Ubuntu Jammy, as Noble also has python 3.12?

Currently we have baseimages for Alpine 3.19-3.22 and edge that are maintained. All except 3.19 have python 3.12. My guess is, Alpine 3.24 will have python 3.13 but that won't be out for another 6-8 months.

txtsd:
If we use `uv`, it can fetch its own suggested/required python.

barbequesauce (approved):

PR #1735, opened 2025-10-22
Fix: Sort "have" by percent amount on ManageComics page

### LINKED ISSUE #1721: Manage Comics: Have Sort order incorrect
**Description**
When clicking Manage, Manage Comics then sort by 'Have' the list appears to sort by the first digit only, ignoring the proportion of the comics you have. For example:
0/1
0/6
1/1
1/3
1/4
10/10
193/193
2/2
Whereas the sort order should be the proportion of which you have (as with the main list sort order):
0/6
0/1
1/4
1/3
1/1
2/2
10/10
193/193
Here's a screenshot to illustrate the problem

Image

**CarePackage**

[carepackage_276d50f33d2ae46fdfee89a5e4d92b0c75fd828f_(python3-dev).zip](https://github.com/user-attachments/files/22059000/carepackage_276d50f33d2ae46fdfee89a5e4d92b0c75fd828f_.python3-dev.zip)

**Environment:**
 - OS: Windows10
 - Branch / commit: Dev/276d50f3

**Additional information**
I have replicated this on multiple browsers and on both my primary and backup machine.
I have also noticed that if I select comics and refresh and any of those comics no longer exist on comicvine that Mylar stops refreshing rather than skipping to the next comic.

barrelltitor:
Fixes #1721

Currently the manage comics page does string sorting on the field, so if you have 194/242 and 5/6 and 1/1 you will get the sorting
1/1 192/242/5/6 

The sorting order should be based on the percent, not on string comparison. Simple change

### DIFF
--- a/data/interfaces/default/managecomics.html
+++ b/data/interfaces/default/managecomics.html
@@ -88,7 +88,7 @@
                                 <td class="hidden" id="stat_title">${comic['Status']}</td>
 				<td id="latest">${comic['LatestIssue']} (${comic['LatestDate']})</td>
                                 <td id="publisher">${comic['ComicPublisher']}</td>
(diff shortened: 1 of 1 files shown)


barrelltitor:
image
Tested on my main install and works great

PR #1739, opened 2025-11-04
fixed image fetching issue

barrelltitor:
image
If we leave the user-agent to the default comictagger one for image fetching it works just fine


kalinon:
yeah this is needed

PR #1741, opened 2025-11-15
Python 3.12 compatibility

barrelltitor:
- Removed unused imghdr which prevented compatibility
- Need to search for any other incompatibility issues

### DIFF
--- a/mylar/PostProcessor.py
+++ b/mylar/PostProcessor.py
@@ -481,7 +481,7 @@ class PostProcessor(object):
                 manual_list = []
                 for fl in filelist['comiclist']:
                     if all([fl['series_name'] is not None, fl['series_name'] != '']) and mylar.CONFIG.IGNORE_COVERS is True:
-                        cvchk = re.sub('[\s\s+\_\.]', '', fl['series_name']).lower()
+                        cvchk = re.sub(r'[\s\s+\_\.]', '', fl['series_name']).lower()
                         if any(['coveronly' in cvchk, 'coversonly' in cvchk]):
                             logger.fdebug('Cover only detected. Ignoring result.')
                             continue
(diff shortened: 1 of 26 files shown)


barrelltitor:
- Replaced single quote strings in re.sub with r strings, just added r before the single quotes. This is needed in python3.12's `re` module
This can be tested using the image 
`ghcr.io/barrelltitor/mylar3:312`

txtsd (approved):

falo2k:
Skimming the diffs quickly:
- Highlighted a possible existing issue in filers.py - line 589, \* isn't an escaped character in python so that's probably never been working as expected :)  Probably only stood out because it wasn't part of a regex change.  Maybe a review of string escapes is needed at some point / I may get there eventually building out tests.
- More seriously, there're a bunch of (unintended?) indentation changes in rsscheck.py.

barrelltitor:
Forgot to reply here, but my last force push fixed the issues above

I don't know what to do about those escapes as I'm not sure what they're meant to do, I would wait for EH's input

mylar/PostProcessor.py
```diff
@@ -790,10 +790,10 @@ def Process(self):
 
                     if not comicseries or orig_seriesname != mod_seriesname:
                         if any(['special' in orig_seriesname.lower(), 'annual' in orig_seriesname.lower()]) and all([mylar.CONFIG.ANNUALS_ON, orig_seriesname != mod_seriesname]):
-                            if not any(re.sub('[\|\s]', '', orig_seriesname).lower() == x for x in loopchk):
-                                loopchk.append(re.sub('[\|\s]', '', orig_seriesname.lower()))
-                                tmpsql = "SELECT * FROM comics WHERE DynamicComicName IN ({seq}) COLLATE NOCASE".format(seq=','.join('?' * len(loopchk)))
-                                comicseries = myDB.select(tmpsql, tuple(loopchk))
+                            if not any(re.sub(r'[\|\s]', '', orig_seriesname).lower() == x for x in loopchk):
+                                loopchk.append(re.sub(r'[\|\s]', '', orig_seriesname.lower()))
+                            tmpsql = "SELECT * FROM comics WHERE DynamicComicName IN ({seq}) COLLATE NOCASE".format(seq=','.join('?' * len(loopchk)))
```
falo2k, line 795: Just doing a quick sanity check, was this change in indent level intentional for this and the next line?

barrelltitor: I figured it out. This was indeed intended, bug fix. This is a duplicate guard.

falo2k (approved):
Sorry, missed this notification somehow!  All of these changes seem like no brainers to me - even if we weren't pushing for a new version, going through the regex matchers to convert to raw strings is a good thing (and a painful bit of work to validate) so has a lot of value.  We talked about that indent change and I'm happy with the explanation.

PR #1745, opened 2025-12-05
Merge suggestion

falo2k:
Thought I would bundle together some changes that I think are fairly important (fixing issues that make noise in #support), and pretty clear cut so could go straight into nightly then push through to master.  Combines and cleans up the following PRs with a rebase on current p3-dev: #1739, #1735, #1733, #1732

On the subject of a push to master, might be time to do one?  The biggest new change in the current nightly is probably the airdcpp support.  A number of people have been using it to a degree of success, so I think it's safe to put in the wild, though it could do with being followed up with:
- Some documentation / user guide
- A solution for post-process move/copy/link options for DC++ given the desire to avoid file manipulation.  Not sure whether that should be a move to having a setting for every source, or just a specific override setting for airdcpp (&torrents?).  Leaning towards the latter, partly to highlight it to end users as a key consideration.

That would in turn enable pushing @barrelltitor's JD2 change into nightly to get some broader testing on it (+some other changes).

### DIFF
--- a/data/interfaces/default/manage.html
+++ b/data/interfaces/default/manage.html
@@ -447,6 +447,7 @@
                         {
                                 $("#move_options").show();
                                 $("#path_options").hide();
+                                $("#imp_paths").prop("checked", false);
                         }
                 else
                         {
@@ -459,6 +460,7 @@
                         {
                                 $("#move_options").slideDown();
                                 $("#path_options").slideUp();
+                                $("#imp_paths").prop("checked", false);
                         }
                         else
                         {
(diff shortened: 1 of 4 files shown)


barrelltitor:
I'd like them to stay separate prs I dont agree personally with this organizational way

falo2k:
Was just trying to make sure they all commit without merge conflicts to minimise downstream effort.  In this case they did, but github can be surprisingly awkward with auto-merging what seem like simple non-conflicting merges from individual PRs.  The commit history afterwards still remains the same.

barrelltitor:
I misunderstood the point of the PR, my apologies 🙏 
I actually do think it is a great idea to get some of the smaller stuff merged in together and quicker, all of the PRs added here are important fixes and should be added as soon as we can imo

barrelltitor:
I made a test image using the LSIO Dockerfile for these changes
`ghcr.io/barrelltitor/mylar3:falos_changes`
It is a drop in replacement based on [this branch](https://github.com/barrelltitor/mylar3/tree/falos_changes) which just merges in this pr

In hindsight, I should have just pulled falo's branch directly and build from there

If you need a fix from here and want to use it, as in any change to a development build, please backup beforehand

barbequesauce (approved):

PR #1749, opened 2025-12-11
Add `pyproject.toml`

txtsd:
This enables the use of the excellent tool [uv](https://docs.astral.sh/uv/) to manage packages, python versions, and the development environment.

Just `uv run Mylar.py`, it will get the pinned python version, install all dependencies, and launch mylar3.

This is intended to be the initial change required to turn mylar3 into a `pip install`able package (which requires some directory restructuring). This will use a build system to turn the code into a wheel.

Looking forward to feedback!


barbequesauce:
@txtsd curious what spurred this idea - what other related "media acquisition ecosystem agents" have you see using uv ?

Issue #1751, opened 2025-12-28
Pull List empty (2026 week 0)


amatriain:
When I open the pull list for December 28, 2025 - January 03, 2026 the list is empty.  Recreating or refreshing the pull list doesn't populate it.

**Debug log:**

Log when recreating the pull list: 

```
28-Dec-2025 12:08:37 - DEBUG :: mylar.toggleVerbose.5827 : CP Server Thread-11 : If you can read this message, debug logging is now working

28-Dec-2025 12:08:37 - INFO :: mylar.toggleVerbose.5830 : CP Server Thread-11 : config file has been loaded from: /config/mylar/config.ini

28-Dec-2025 12:09:06 - INFO :: mylar.pullist.2930 : CP Server Thread-10 : trying to repopulate to week: 0-2026

28-Dec-2025 12:09:06 - INFO :: mylar.pullrecreate.3153 : CP Server Thread-10 : [PULL-LIST] Now Recreating the Pull-list for week 0, 2026...

28-Dec-2025 12:09:06 - INFO :: mylar.pullit.72 : CP Server Thread-10 : [PULL-LIST] Populating & Loading pull-list data directly from alternate website

28-Dec-2025 12:09:08 - INFO :: mylar.pullit.98 : CP Server Thread-10 : [PULL-LIST] Populating & Loading pull-list data directly from alternate website for specific week of 0, 2026

28-Dec-2025 12:09:09 - WARNING :: mylar.locg.64 : CP Server Thread-10 : [619] No date supplied, or an invalid date was provided [None]

28-Dec-2025 12:09:09 - WARNING :: mylar.pullit.112 : CP Server Thread-10 : [PULL-LIST] Unable to retrieve weekly pull-list. Pull list for week 0, 2026 may be stale.

28-Dec-2025 12:09:09 - WARNING :: mylar.pullist.2935 : CP Server Thread-10 : Problem repopulating the pullist for week 0, 2026

28-Dec-2025 12:09:09 - WARNING :: mylar.pullist.2937 : CP Server Thread-10 : Attempting to repoll against legacy pullist in order to have some kind of updated listing for the week.

28-Dec-2025 12:09:09 - INFO :: mylar.manualpull.3132 : CP Server Thread-10 : ALT_PULL: 2 PULLBYFILE: False week: None year: None

[...]

PR #1755, opened 2026-01-05
Fix 2026 week rollover for pull list

jfv514:
## Summary
- Ensure week/year are coerced to ints before date math in weekly_info()
- Handle 2025/2026 rollover to avoid week 0 for the prior-week calculation

## Context
Fixes pull list empty issue in week 0 of 2026 (see #1751).

## Testing
- Verified on a Mylar3 docker deployment: /upcoming and /pullist no longer error during week rollover.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -3155,6 +3155,9 @@ def weekly_info(week=None, year=None, current=None):
         weeknumber = 51
         year = 2024
 
+    # normalize to ints before date math to avoid weeknumber being a string
+    weeknumber = int(weeknumber)
+    year = int(year)
     startofyear = date(year,1,1)
     week0 = startofyear - timedelta(days=startofyear.isoweekday())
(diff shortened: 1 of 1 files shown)


jfv514:
Repro: week 1 of 2026 triggers week 0 in weekly pull list (empty pull list + errors in helpers.weekly_info when weeknumber is a string). This patch coerces week/year to int before date math and handles the 2025→2026 rollover to avoid week 0. Tested on Docker: /pullist and /upcoming OK after change.

falo2k:
- Pulls go against the dev branch
- A patch for the issue introduced by the 2026 rollover causing a type error has already been merged
- No changes have been made to the week logic as yet as that is not the only problem with the missing week.  Without an update to the backend you are still missing the actual week of data, and will just realign the wrong data to the wrong week.

jfv514:
Thanks for the clarification. Understood about targeting python3-dev and the backend/data source issue. We’ll keep this as a local workaround only and I’ll close the PR.

falo2k:
np.  Thanks for taking a look at it anyway, the effort is appreciated.

PR #1764, opened 2026-01-30
Add an optional paging parameter to the findComic API endpoint

dangle:
This is a bit hacky, but I wanted to minimize any sort of alterations to the default use case.

### DIFF
--- a/API_REFERENCE
+++ b/API_REFERENCE
@@ -63,7 +63,7 @@ restart (restart mylar)
 update (update mylar - you may want to check the install type in get version and not allow this if type==exe)
 
 downloadIssue&id=$issueid (download that issue to your browser)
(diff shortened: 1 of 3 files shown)


falo2k:
Is there a reason you slice the results array for paging at the end, rather than when the search query is returned?  Feels like that's leaving a lot of wasted processing on the table.

dangle:
I was trying to minimize alterations to the actual processing loop. I don't mind changing the code, but I think this is fairly minimal waste, considering it only pulls, at most, one page of data. This really only impacts the very last page pulled if it has less than `pageSize` data.

barbequesauce (changes requested):
Please resubmit against the 1000papercuts branch.

PR #1790, opened 2026-02-12
IMP: Add Easynews as a DDL provider for direct comic downloads

FrankEBailey:
## Summary
Adds [Easynews](https://www.easynews.com/) as a Direct Download (DDL) provider alongside GetComics and External Server. Easynews decodes binaries server-side and provides a JSON search API with direct HTTP file downloads using Basic Auth, so it fits into the existing DDL flow without extra client-side decoding.

## Changes
- **New module** `mylar/easynews.py`: search (Solr-style API), result normalization to the existing DDL format, and streaming download with Basic Auth. Supports CBR/CBZ/CB7/PDF; handles ZIP-wrapped results.
- **Config** (`mylar/config.py`): `enable_easynews`, `easynews_username`, `easynews_password`; DDL(Easynews) in provider order and `write_out_provider_searches` (provider id 202).
- **Search** (`mylar/search.py`): Easynews in DDL provider list and search path; results go through the same filer and DDL queue as other DDL providers.
- **Download path** (`mylar/helpers.py`, `mylar/search_filer.py`): DDL(Easynews) snatches use the existing DDL worker and post-processing.
- **Web UI** (`mylar/webserve.py`, `data/interfaces/default/config.html`): Easynews options on the config page and a “Test Easynews” connection button.

## Usage
1. Enable DDL and Easynews in Config → Search Providers.
2. Set Easynews username and password (member credentials).
3. Use “Test Easynews” to verify; optionally adjust DDL provider order.

No new dependencies; uses existing `requests` and proxy settings.

### DIFF
--- a/data/interfaces/default/config.html
+++ b/data/interfaces/default/config.html
@@ -868,6 +868,23 @@
                                           </div>
                                        </div>
                                        %endif
+                                       <div class="row checkbox left clearfix">
+                                            <input type="checkbox" id="enable_easynews" onclick="initConfigCheckbox($(this));" name="enable_easynews" value="1" ${config['enable_easynews']} /><label>Enable Easynews</label>
+                                       </div>
+                                       <div id="easynews_options" style="display:none">
+                                           <div class="row">
+                                              <label>Username</label>
+                                              <input type="text" name="easynews_username" id="easynews_username" value="${config['easynews_username']}" size="30">
+                                           </div>
+                                           <div class="row">
+                                              <label>Password</label>
+                                              <input type="password" name="easynews_password" id="easynews_password" value="${config['easynews_password']}" size="30">
+                                           </div>
+                                           <div align="center" class="row">
+                                              <input type="button" value="Test Easynews" id="test_easynews" />
+                                              <input type="text" name="easynewsstatus" id="easynewsstatus" style="text-align:center;font-size:11px;color:white;" size="50" DISABLED />
+                                           </div>
+                                       </div>
                                    </div>
                                 </div>
                         </fieldset>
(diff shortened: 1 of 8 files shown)


FrankEBailey:
I'm closing this for now, the cleanup in preparation for the PR caused some regressions.

barrelltitor:
This also needs to be targeted on 1000papercuts, not master

falo2k:
I'd also suggest popping into the discord to discuss this change.  Among a few technical changes I'd want to see, I'm struggling to see the value of adding this.  As I understand it, it's a very niche service.  Does it even get decent releases (given we know where most of the usenet releases originate from) and can it handle obfuscated ones?

FrankEBailey:
>  can it handle obfuscated ones?

Obfuscated in what way?

coreyemtp:
I'm not sure why one wouldn't just add Easynews to SabNZBD as a server. I have it this way right now. While they do provide the web interface, it really has nothing to do with the underlying service. Adding it as a DDL setting that not everyone could use would be redundant and potentially confusing IMO.

FrankEBailey:
I'll quite happily putter along on my own fork then, adding all sorts of Byzantine nonsense nobody'll ever make use of but myself :D

FrankEBailey:
>While they do provide the web interface, it really has nothing to do with the underlying service. Adding it as a DDL setting that not everyone could use would be redundant and potentially confusing IMO.

The differentiator with Easynews is that they decode the content themselves and serve it as straight HTTP downloads on 443. And they have like 40 years of decoded binaries so their depth of content is enormous, and their search is very well-optimised and feature-rich so it enables tuning of searches, exclusion of terms and a few other unique features.

That said, it's definitely niche, so I'll go play in my own sandbox, no harm no foul.

PR #1791, opened 2026-02-12
FIX: exception adding new series

jramboz:
There was an extra `\` in the regex, causing the final `]` to be escaped. As a result, adding new series was causing an exception.


barrelltitor (approved):
Good catch

barbequesauce (approved):

PR #1793, opened 2026-02-13
FIX: weekly_info() TypeError for years after 2025

### LINKED ISSUE #1753: 500 server error when viewing Wanted and This Week
Currently receiving a 500 Server error when going to Wanted or This Week. I've attached my CarePackage. Not exactly sure when this started. I had previously noticed that it wasn't producing a Pull List for Week 0. I came back a few days later to check if a Pull List was available and began seeing the 500. 

[carepackage_v0.8.3.zip](https://github.com/user-attachments/files/24400862/carepackage_v0.8.3.zip)

 - OS: Docker
 - Branch / commit: master

The traceback mentions TypeError: unsupported type for timedelta weeks component: str

bitr8:
Fixes #1753.

`strftime("%U")` returns a string, but `timedelta(weeks=...)` needs an int. The 2025 branch already uses `isocalendar()[1]` which returns int, so this was only broken for other years. Wrapping in `int()` fixes the 500 on Wanted/This Week pages.

Test expectations updated to match.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -2704,7 +2704,7 @@ def weekly_info(week=None, year=None, current=None):
     if todaydate.year == 2025:
         current_weeknumber = todaydate.isocalendar()[1]
     else:
-        current_weeknumber = todaydate.strftime("%U")
+        current_weeknumber = int(todaydate.strftime("%U"))
     if current is not None:
         c_weeknumber = int(current[:current.find('-')])
         c_weekyear = int(current[current.find('-')+1:])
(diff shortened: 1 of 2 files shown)


falo2k:
Hi.  This has already been fixed in python3-dev.  Further changes to that function aren't necessary, and need to be done in the context of an update to the WS backend.

Future PRs should be done against the 1000papercuts branch as we are currently using that as the staging for development.  I'd suggest jumping on the discord where there's more visibility of ongoing activity.

bitr8:
Cheers, missed that it was already fixed. Will target 1000papercuts for future PRs.

PR #1794, opened 2026-02-13
FIX: Reset issue status to Wanted when DDL download exhausts all links

bitr8:
DDL downloads that exhaust all links get stuck in Snatched permanently, never retried via usenet/torrent.

Three problems in `ddl_downloader`:

1. GetComics exhausted-links path never reset the individual issue to Wanted (only called `reverse_the_pack_snatch`).
2. External DDL path called `FailedMark()` *after* setting Wanted, which clobbered it back to Failed. Reordered so the Wanted upsert wins.
3. External DDL path was missing `ddl_cleanup()` and `DDL_QUEUED.remove()`, so failed items stayed in the queue forever.

Also added `try/except ValueError` on `DDL_QUEUED.remove()` since the success path may have already cleared it.

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -3033,14 +3033,34 @@ def ddl_downloader(queue):
                         nval = {'status':  'Failed',
                                 'updated_date': datetime.datetime.now().strftime('%Y-%m-%d %H:%M')}
                         myDB.upsert('ddl_info', nval, ctrlval)
-                        #undo all snatched items, to previous status via item['id'] - this will be set to Skipped currently regardless of previous status
+                        # undo all snatched items, to previous status via item['id']
+                        # pack siblings are set to Skipped (opportunistic); individual issue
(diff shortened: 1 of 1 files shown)


falo2k:
Please resubmit this against the 1000papercuts branch.  

The Failed status should still be added to the search list when auto-search runs.

I'm not particularly comfortable with catching and just passing exceptions.

bitr8:
Going to rework this based on your feedback. Will resubmit against 1000papercuts once I've got the Failed status flow right.

falo2k:
> Going to rework this based on your feedback. Will resubmit against 1000papercuts once I've got the Failed status flow right.

Awesome, cheers.  It's definitely one of those states that's bugged me occasionally so would be good to see it fixed.  Sorry it was a bit of a cursory scan for that feedback - will give it a proper look on the resubmit, but it looks broadly sensible :)

PR #1796, opened 2026-02-14
Feature/bookshelf view

FrankEBailey:
A simple bookshelf function for the eye candy enjoyers. No additional images are downloaded, no additional ComicVine API calls are made, all cover images are cached in the browser.

### DIFF
--- a/data/interfaces/default/comicdetails_update.html
+++ b/data/interfaces/default/comicdetails_update.html
@@ -639,7 +639,6 @@
           background: rgba(0,0,0,0.2);
           border-radius: 6px;
           overflow: hidden;
-          cursor: pointer;
           min-width: 0;
         }
         .bookshelf-card img {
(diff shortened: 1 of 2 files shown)


falo2k:
Right, I've had a play with this and I'm not personally keen on this feature or implementation.
- My headline concern is still the same.  I don't really see the point in this for a downloader/organiser.  The space has moved on to a point where reading servers (some very lightweight like codex) exist and do this much better.  This feels like feature for the sake of feature.
- I don't like the styles just shoved into the page like that rather than into style sheets.
- The choice of making it a tab in that view is odd.  It completely breaks the layout of the page.  Surely, as an alternative view on issues, it should swap out with the issues list itself.  In addition, it doesn't respect any of the existing issue visibility controls (page size, filters).  Feels like it's just slapped in there.
- It's just linking directly to the comicvine hosted images.  We've had enough problems with comicvine changing their cloudflare settings in recent months, and I don't fancy sparking any more.  It also means that the covers being shown aren't representative of your files.  Given all that, I really don't see what this gives you over clicking on the volume title on the page, which opens up ComicVine at that volume.

DieselTech:
I'm also of the mind that Mylar doesn't need any feature creep right now while we are scrambling to fix bugs and keep the project moving along in the absence of the original creator. If you like eye candy, then running one of the reading servers might better suit your needs:

 - https://github.com/Kareadita/Kavita
 - https://github.com/ajslater/codex
 - https://github.com/stumpapp/stump
 - https://github.com/gotson/komga
 
In another timeline where things were smooth sailing with the project I would say go for it. Little touches like this are nice and help users enjoy the software more.  But the current state of things doesn't have us in a position of luxury features at the moment.

FrankEBailey:
Thanks for your input, folks.

PR #1754, opened 2026-01-01
FIX: Quick fix for 2026 This Week page loading

falo2k:
Quick fix to resolve This Week not loading after the new year due to a missing cast.  This does not fix the underlying problem of year/week mapping for the new year and WS as that will need to be a unified solution.

To close #1753

PR #1800, opened 2026-02-18
FIX: TypeError in weekly_info() - year-specific weeknumber branching causes str passed to timedelta

### LINKED ISSUE #1753: 500 server error when viewing Wanted and This Week

DeusExTaco:
## Problem

Clicking the **Wanted** (`/upcoming`) or **This Week** (`/pullist`) tabs
produces a 500 Internal Server Error:
```
TypeError: unsupported type for timedelta weeks component: str
```

This affects all users in 2026+ on the master branch.

## Root Cause

In `helpers.py`, the `weekly_info()` function uses a year-specific
conditional (line ~3105) that only returns an integer weeknumber for
2025 via `isocalendar()[1]`. For all other years, it falls through to
`strftime("%U")` which returns a **string**. This string is then passed
to `timedelta(weeks=weeknumber)` which requires an integer.

A second year-gated block (line ~3161) pairs `isocalendar` with a
`weeknumber - 1` offset (since ISO weeks are 1-indexed) but only for
2025, using `weeknumber` (0-indexed) for other years.

## Fix

Remove both year-specific conditionals and use universal logic:
- `isocalendar()[1]` for all years (always returns int, no type error)
- `weeknumber - 1` offset for all years (correct for 1-indexed ISO weeks)

This matches the pattern the developer already established for 2025,
extended to work for every year.

Fixes #1753

### DIFF
--- a/mylar/helpers.py
+++ b/mylar/helpers.py
@@ -3102,10 +3102,7 @@ def torrentinfo(issueid=None, torrent_hash=None, download=False, monitor=False):
 def weekly_info(week=None, year=None, current=None):
     #find the current week and save it as a reference point.
     todaydate = datetime.datetime.today()
-    if todaydate.year == 2025:
-        current_weeknumber = todaydate.isocalendar()[1]
-    else:
-        current_weeknumber = todaydate.strftime("%U")
+    current_weeknumber = int(todaydate.isocalendar()[1])
     if current is not None:
         c_weeknumber = int(current[:current.find('-')])
         c_weekyear = int(current[current.find('-')+1:])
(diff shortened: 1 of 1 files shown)


professionaltart:
Hi @DeusExTaco, I think this one is already fixed in the Python3-dev and 1000Papercuts branches via https://github.com/mylar3/mylar3/pull/1754

There'll be a few other bugfixes and features in there too. - can you please try swapping branch and see if that works for you?  If you're using docker images and not source from Github, Python3-dev correlates to :nightly, and 1000Papercuts to :unstable

Going forward, PRs are being accepted against the 1000papercuts branch.

DeusExTaco:
@professionaltart Sorry about that I changed branches and it is fixed. Thanks for the heads up on the right branch for PRs.

professionaltart:
All good - we appreciate people wanting to contribute with PRs, but we're in a temporary bind where we can't push updates to master/latest, which means we also can't add a note to the default readme that loads for the project.

DeusExTaco:
Closing — confirmed fixed in #1754 on the 1000papercuts branch. Will submit future PRs against that branch.

Issue #1704, opened 2025-08-20
Issue Matching StoryArc to Annuals when Annual Isn't Related to Series (and Annual Integration is On)


caddy-wampus:
**Describe the bug**
When all of these conditions are met:
1. You have annual integration enabled
2. An annual is not directly related to a series (either the name doesn't match the series it is connected to or it is just a one-off)
3. An annual is part of a story arc

In this scenario, Mylar will not be able to match the issue as part of the story arc.

**Example**
Ms. Marvel Annual #1 (annual series id 159124). This is part of the Infinity Watch story arc (id 61216). The closest Ms. Marvel series is Ms. Marvel Mutant Menace (series id 157125).

**Debugging**
I attached a debugger and stepped through what happens when I click the Search for Watchlist/Recheck Files button. (this happens in the ArcWatchList function in webserve.py)
1. It looks up the comic from the comics table. Because annual integration is on, it strips out "annual" from the name, resulting in it matching the wrong comic.
  a. In the example above, it will match the closest "Ms. Marvel" series rather than the "Ms. Marvel Mutant Menace" series if you have attached it to that.
  b. If you instead add the annual as its own series, it will still match the "Ms. Marvel" series instead of the "Ms. Marvel Annual" series.
2. Because annual integration is ion, for the comics it finds in #1, it tries to look up the annual from that comic. In the example above, it is looking at the wrong comic so it won't find anything.

**Environment (please complete the following information):**
 - OS: Docker container / Linux (reproduced on both)
 - Branch / commit: cdc94a44

**Additional information**
Feel free to assign this issue to me and I will work on a fix.

My proposal would be to use the ComicID field from the StoryArc table to look the issue up instead of doing it by name. That way we can match either an annual integrated with a series or an independent annual added as a series of its own.

Issue #1801, opened 2026-02-19
Having issues with Mylar?  Read this first!


professionaltart:
While we are temporarily unable to release new versions or push changes to Mylar's main branch, we'd like to make it clear that **Mylar is still an active project**, with bugfixes (and features) being added to to the development branches.  

As it currently stands, there are three branches:
- [master](https://github.com/mylar3/mylar3/tree/master) (github/source) / ~~:latest~~ _(this branch is currently not available as a  docker image after a change to the CI build chain)_ 
  - traditionally the stable/release branch
  - not currently being updated, no longer stable
- [python3-dev](https://github.com/mylar3/mylar3/tree/python3-dev) /~~:nightly~~ **:latest** (_[docker image](https://hub.docker.com/r/linuxserver/mylar3) CI build chain updated, this is now deployed as the :latest image_)
  - traditionally the (mostly stable) development branch
  - now the active stable branch until we can push to master again
- [1000papercuts](https://github.com/mylar3/mylar3/tree/1000papercuts) / **:nightly** (ubuntu), **:unstable** (alpine) (_shuffled after above-mentioned CI changes_)
  - a new branch for testing before changes are pushed to python3-dev
  - any new PRs should be made against this branch

There are a number of known bugs with the master branch's v0.8.3 release.  The most obvious, if you haven't opened Mylar for a little while, is a 500 server error when clicking on **Wanted** and **This week** to see your current /pullist.  This has been patched in the development branches.

For now, we recommend you update Mylar to a more recent branch.  If you're on **master**, swap to **python3-dev**.  If you're on :latest, ~~swap to :nightly~~ you can now stay there.

If you are still seeing bugs after those updates, please log a new [Issue ](https://github.com/mylar3/mylar3/issues) on here, or jump onto [Discord](https://discord.gg/6UG94R7E8T), [IRC](https://web.libera.chat/?channels=#mylar), or the [forums](https://forum.mylarcomics.com/viewforum.php?f=8)

If you'd like to make a code submission, we welcome contributions! At present we ask that you fork and submit changes to the 1000papercuts branch. The python3-dev branch is considered stable and we will merge 1000papercuts to it on a regular basis once given changes have been proven.  If you are making a non-trivial change, <ins>please jump on Discord first to discuss it with the development team</ins> to make sure it won't conflict with any other work in the pipeline.

[...]

PR #1810, opened 2026-03-06
Check story arc issues by ID, to fix annual integration issues

caddy-wampus:
When checking for issues in a story arc, first check by id within the DB. This is faster and more reliable in cases where things like annuals may not line up exactly. I kept the fallback to the existing behavior if it is not found by ID.

I tested this on story arcs where annuals are cleanly tied to their parent series, and story arcs where the annual was tied to a differently named parent series. I also tested story arcs where some annuals were added directly as series rather than tied to the series. With this fix, all of these cases worked. Without this fix, only annuals that matched the parent comic's name and were added as annuals would match.

I also tested with annual integration turned off, and this behaved as expected.

#1704  was marked as not planned, but this is a much delayed fix for that issue.

### DIFF
--- a/mylar/webserve.py
+++ b/mylar/webserve.py
@@ -5386,6 +5386,50 @@ class WebInterface(object):
 
                 matcheroso = "no"
 
+                # First try matching by IDs in the existing DB
+                if mylar.CONFIG.ANNUALS_ON:
+                    comics_by_id = myDB.select(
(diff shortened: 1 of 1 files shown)


falo2k:
Please resubmit this against 1000papercuts.  See #1801

caddy-wampus:
Sorry I missed that - I have updated to the correct branch.

caddy-wampus:
I created a PR on the new repo for this change.

PR #1813, opened 2026-03-08
Optionally use curl_cffi instead of requests

jramboz:
Add an option to use `curl_cffi` instead of `requests`. Option is enabled by default.

`curl_cffi` has built in anti-bot evasion, which should make Flaresolverr unnecessary in most cases. It does use a binary, which means it may not be available for all platforms; hence why it's an option.

### DIFF
--- a/mylar/config.py
+++ b/mylar/config.py
@@ -380,6 +380,7 @@ _CONFIG_DEFINITIONS = OrderedDict({
     'JD2_ENABLE': (bool, 'DDL', False),
     'JD2_DEST_DIR': (str, 'DDL', None),
     'JD2_URL': (str, 'DDL', None),
+    'USE_CURL_CFFI': (bool, 'DDL', True),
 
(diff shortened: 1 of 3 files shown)


falo2k:
You've added the option, but then removed the import for requests.  It's also not really an optional change if you make it default to on, and make it a config file only option.  This will cause all sorts of noise with source installs who are notoriously bad at env management.

This, from the pypi page for that library, does not fill me with confidence.
```
This should work for Linux(x86_64/aarch64), macOS(Intel), Windows(amd64). If it does not work, you may need to compile and install curl-impersonate first.
```

jramboz:
I removed one of the includes for `requests`. It was included twice. It's still being included and used if the option is disabled. It's included at line 18, and used at line 141.

jramboz:
I've tested native installs now on Windows (10 and 11), macOS, Linux (Alpine), and Docker, all using Python 3.14. All installed without error and worked as expected.

PR #1827, opened 2026-04-06
AirDC++ Download update.

Cducharme84:
Current implementation passes the AirDC++ download location to AirDC++ manually, this causes pathing issues with Windows installs of AirDC (see discord conversation in #support on April 6th 2026). The two payload attributes `target_name` and `target_directory` are optional and default to using `filename` and `dl_location` by default. With the translation via API of `/` to `_` this results in the first image below as filename and attempt to create an invalid file on the windows install.
 
image

By leaving these payload items out, we're allowing AirDC to manage where the files go within its system and to pass along the resulting file to Mylar where we already account for the system path differences in the function `download` on what is now line `391`.

Reference docs/endpoint screenshot for easy verification.
[API Doc on endpoint](https://airdcpp.docs.apiary.io/#reference/search-instances/results/download-result)
Screenshot_20260406_102443

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,11 @@ It will also allow you to monitor weekly pull-lists for items belonging to said 
 Install it via git clone or via [Docker](https://hub.docker.com/r/linuxserver/mylar3)
 
 ## Documentation
-Check out our [website](https://mylarcomics.com) for documentation!
+Check out our [website](https://mylar.nerdfirehurricane.com) for documentation!
 
 ## Support & Discussion
 Please try to limit Github issues to bugs & enhancement requests ONLY
 - [Github](https://github.com/mylar3/mylar3/issues) (Bug & Feature requests only)
-- [Forums](https://forum.mylarcomics.com)
 
 ## Development
 We welcome contributions! At present we ask that you submit changes to the 1000papercuts branch. 
@@ -28,11 +27,11 @@ basis once given changes have been proven.
 
 ## Live Support / Conversation
 - [Discord](https://discord.gg/6UG94R7E8T)
-- [IRC](https://web.libera.chat/?channels=#mylar)
 
 ## Features
 - Abliity to be run on various OS' (windows, linux, macOS, Raspberry Pi, etc)
 - Support for SABnzbd, NZBGet and various torrent clients (as well as Blackhole)
+- AirDC++ Support for pulling releases from your favorite hubs
 - Multiple newznabs support, as well including a raw indexer and direct download being available
 - Ability to see upcoming new releases for a particular week and take action on them if required
 - View pullists up to 4 weeks in advance, or several months prior
@@ -50,7 +49,7 @@ basis once given changes have been proven.
 AND SO MUCH MORE!
 
 ## Contributing
-If you wish to help out, please see our website: [contributing](https://mylarcomics.com/docs/contributing)
+If you wish to help out, please see our website: [contributing](https://mylar.nerdfirehurricane.com/docs/contributing)
  
-<p align="center">This project exists thanks to all the people who contribute - whether by code, assisting others or financial donations.</br> 
+<p align="center">This project exists thanks to all the people who contribute - whether by code, or assisting others</br> 
 To all those who have contributed, we thank you for your support.</p>
--- a/mylar/downloaders/airdcpp.py
+++ b/mylar/downloaders/airdcpp.py
@@ -404,8 +404,6 @@ class AirDCPP(object):
                 search_instance_id = self.current_search_instance_id
 
             payload = {
-                "target_directory": dl_location,
-                "target_name": filename,
                 "priority": 4
             }
 


Cducharme84:
Confirmed to resolve the issue for user shadoweave and allow airdc to grab download, and with a properly filled in download path for the mylar config allowed post-process.

Personal to-do in future PRs:
1. Default blank download location to provide warning that it is encouraged to always put in the location as mylar sees it, also update to have blank entry attempt `target` from [response payload](https://airdcpp.docs.apiary.io/#reference/queue/get-bundles) as a directory before failing post-processing as file not found.

2. High priority - Copy option for AirDC++ specifically, allowing a hash perfect copy to remain in share while post-processing for other downloaders are unaffected.

shadoweave:
Just want to update that this edit works great. But it requires a properly filled in download path otherwise Mylar will try to fetch from a default /config/mylar/cache/airdcpp/ folder.

Cducharme84:
> Just want to update that this edit works great. But it requires a properly filled in download path otherwise Mylar will try to fetch from a default /config/mylar/cache/airdcpp/ folder.

Yep, I have thoughts on how to avoid this. The response from Air's API includes a `target` that has the path as Air sees it. Way I reckon the situation is we can assume if the location in mylar's config is blank we want to try that `target` path first, if failed THEN put in the logs an error warning telling the folks to update that config option. Otherwise we can convert from `target` using the data in the field.

I'm currently looking into whether we're even querying the best endpoint, right now it's interacting with `/queue/bundles` which checks everything in the queue vs the targeted `/queue/files/` with `tth` as the argument which will return just one result including full path.

The reason I think we should use `files` is because this returns a full path and will allows us to recalibrate for a differing named file (as Air matches by hash sometimes when you add to the queue the file name in primary position has changed causing the current logic to shit itself.)

I'm looking a bit more into today, I'm confident on the API change and as a non-AI generation user with only light (And thus double and triple checked) python coding knowledge I hope to have time to sit down and build out my few changes to this flow in the next week or two.

falo2k:
Given the above, I would say it's reasonable to resume passing filename to adcpp, and use a filename that embeds a reference ID (e.g. the issue ID, or preferably some unique identifier for that download session).  As you then know that filename, you should then be able to determine the root path below it on the return leg and swap that for the provided mapped path on the mylar side.

shadoweave:
> Yep, I have thoughts on how to avoid this. The response from Air's API includes a `target` that has the path as Air sees it. Way I reckon the situation is we can assume if the location in mylar's config is blank we want to try that `target` path first, if failed THEN put in the logs an error warning telling the folks to update that config option. Otherwise we can convert from `target` using the data in the field.
> 

I think the issue in this case given the present edge-case of Windows vs Linux is that the AirDC++ target path won't convert well to the Linux path. Especially given that there's no remote path/local path mapping like how Radarr/Sonarr does it.

Cducharme84:
> > Yep, I have thoughts on how to avoid this. The response from Air's API includes a `target` that has the path as Air sees it. Way I reckon the situation is we can assume if the location in mylar's config is blank we want to try that `target` path first, if failed THEN put in the logs an error warning telling the folks to update that config option. Otherwise we can convert from `target` using the data in the field.
> 
> I think the issue in this case given the present edge-case of Windows vs Linux is that the AirDC++ target path won't convert well to the Linux path. Especially given that there's no remote path/local path mapping like how Radarr/Sonarr does it.

That's what the download location field would be doing in my thought process, acting as a path mapper if needed, I was just explaining logic should it be empty. It will certainly fail on a linux to windows as you saw, BUT with linux to linux bare metal to bare metal someone wouldn't need to do a mapping and would presumably make it blank. I was looking for all the potential scenarios rather than fixating on just the present one. 

As it is now, using mylar's cache is just 100% of the time wrong, whereas `target` attempt has a more than zero percent shot :D

falo2k:
> BUT with linux to linux bare metal to bare metal someone wouldn't need to do a mapping and would presumably make it blank

Don't think of it as a mapping, even if it is.  Make it mandatory to point to the airDCPP download root.  You do path substitution regardless based on filename sent and returned full path.  That you may be replacing the same string for the same string becomes irrelevant.  Sprinkle in some additional checks for path separator types.

Cducharme84:
> Given the above, I would say it's reasonable to resume passing filename to adcpp, and use a filename that embeds a reference ID (e.g. the issue ID, or preferably some unique identifier for that download session). As you then know that filename, you should then be able to determine the root path below it on the return leg and swap that for the provided mapped path on the mylar side.

With the way air handles queued downloads the filename we capture early in the code on search can be incorrect to what's present (I'm talking pure remote file system discovery here not storage for post processing, naturally we can use the filename obtained earlier and append issueid per post-process normal flow we're just talking before this point) based on this scenario:

1. Mylar queries, creating a search session and hitting the hubs with the search, while waiting mylar's not actively seeing what the search is doing until it checks in at a predetermined time (I haven't looked to confirm but I believe up to 30 seconds)
2. Mylar checks in to see results list, parses and tries to match per logic for issue match, this can take a few seconds to a minute depending on the amount of results
3. During that time, AirDC is still querying the hubs and getting results some of which are the same TTH changing the result's file name as Air shows just an arbitrary "top" result as we see them but keeping result IDs the same
4. Mylar's sending the API call to download that result ID, Air follows its defaults which is use whatever file name is showing there, this can be different than what mylar sees as `filename`
5. Currently there'd be a mismatch as we're looking for an old file name, not what is returned when using the `/queue/files` endpoint causing the issue of possible file name mismatch prior to attempting post-processing and mylar says "where's that dang file?"

Absolutely we can force the download location as a required value, which would remove the need for mapping as you described.

